PR #57, opened 2022-03-11
HTM-77: Zoom to initial extent button

geertplaisier:


mprins (approved):
I don't recall what we agreed we should do about the i18n for (aria) labels? these are English only?  (same for +/- buttons I think?...

mvdstruijk (approved):

PR #60, opened 2022-03-18
HTM-145: Show highlight of currently selected feature info feature 

geertplaisier:
Note: for this to work the geometry attribute needs to be checked in the Attributes tab in 'Boomstructuur met kaarten'
Once HTM-258 is implemented this should no longer be necessary


mprins:
@geertplaisier I've updated FeatureModel, not sure if any other changes are required.

`geometry` attribute should have (simplfied) geometry; but attributes will still have geometry as well if checked in configured attributes list

mprins (approved):

PR #72, opened 2022-03-29
HTM-181 select coordinates

mprins:
HTM-181 describes design etc.


projects/map/src/lib/map-service/map.service.ts
```diff
@@ -158,6 +158,17 @@ export class MapService {
     return this.map.getProjection$().pipe(map(p => p.getCode()));
   }
 
+  /** Gets the UOM of the map as defined in the projection.
+   * Will return 'm' in case the map's projection has not defined UOM (such as EPSG:28992).
```
mprins, line 162: well... either OL does not parse the proj string entirely or something else is going on because the definition we use does have a `units` dimension:
`EPSG:28992[+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.237,50.0087,465.658,-0.406857,0.350733,-1.87035,4.0812 +units=m +no_defs]`

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.destroyed.complete();
+  }
+
+  public toggle() {
+    this.store$.dispatch(toggleTool({ tool: ToolbarComponentEnum.SELECT_COORDINATES }));
+  }
+
+  private handleMapClick(evt: { mapCoordinates: [number, number]; mouseCoordinates: [number, number] }) {
+    let decimals = 4;
+    this.mapService.getUnitsOfMeasure$().subscribe(uom => {
```
geertplaisier, line 60: You should try to avoid making new subscriptions inside subscriptions. This handleMapClick method is called from the context of a subscription. If possible try to chain those streams.

Also this statement is now written as synchrounous statement. You expect the subscription to be done once you reach line 75 but this does not have to be the case since Observables are async.

Lastly, you never unsubscribe here so the subscriptions will keep stacking.

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
```diff
@@ -0,0 +1,85 @@
...
+        case 'degrees':
+        default:
+          decimals = 4;
+      }
+    });
+
+    const coordinatesMsg = $localize`Selected coordinates: ${evt.mapCoordinates[0].toFixed(decimals)}, ${evt.mapCoordinates[1].toFixed(decimals)}`;
+
+    this.snackBar.open(coordinatesMsg, $localize`Copy`).onAction().subscribe(() => {
+      if (this.clipboard.copy(coordinatesMsg)) {
```
geertplaisier, line 78: Maybe just copy the coordinates so they are easier to copy/past in other application without having to remove the 'Selected coordinates' part.

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
```diff
@@ -0,0 +1,85 @@
...
+          break;
+        case 'degrees':
+        default:
+          decimals = 4;
+      }
+    });
+
+    const coordinatesMsg = $localize`Selected coordinates: ${evt.mapCoordinates[0].toFixed(decimals)}, ${evt.mapCoordinates[1].toFixed(decimals)}`;
+
+    this.snackBar.open(coordinatesMsg, $localize`Copy`).onAction().subscribe(() => {
```
geertplaisier, line 77: Again a new subscription, see comment above

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
```diff
@@ -0,0 +1,85 @@
...
+      .pipe(
+        takeUntil(this.destroyed),
+        filter(Boolean),
+        tap(clickTool => {
+          this.store$.dispatch(registerTool({ tool: { id: ToolbarComponentEnum.SELECT_COORDINATES, mapToolId: clickTool.id }}));
+        }),
+        concatMap(clickTool => clickTool.mapClick$),
+      )
+      .subscribe(mapClick => {
+        this.handleMapClick(mapClick);
```
geertplaisier, line 45: See my comments below. Instead of subscribing here and execute this method it would be a better approach to chain the Observables using concatMap/switchMap, so after `concatMap(clickTool => clickTool.mapClick$),` you could add another `concatMap` to convert the coordinates to readable coordinates using the `this.mapService.getUnitsOfMeasure$()` Observable, for example:

```
concatMap(mapClickEvt => {
  return this.mapService.getUnitsOfMeasure$()
    .pipe(map(unit => `${evt.mapCoordinates[0].toFixed(decimals)}, ${evt.mapCoordinates[1].toFixed(decimals)}`));
}),
concatMap(coordinates => {
  const coordinatesMsg = $localize`Selected coordinates: ${coordinates}`;
  return this.snackBar.open(coordinatesMsg, $localize`Copy`).onAction();
}),
```
and then in the subscribe you end up with the 'Copy' button action the user executed so there you could do this code:
```
if (this.clipboard.copy(coordinatesMsg)) {
  this.snackBar.open($localize`Success`, '', {duration: 5000});
} else {
  this.snackBar.open($localize`Failed to copy to clipboard`, '', {duration: 5000});
}
```

geertplaisier (changes requested):

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
mprins: accessibility wise not a good idea, as the message will be announced as just the 2 numbers, also in general most GIS applications receiving a coordinate pair want them as 2 separate values (note that for some CRS they may need to be flipped in order as well). 

It might be better to provide a WKT

mprins:
@geertplaisier Can you take another look?

I have 1 little problem: the tool does not re-activate after the snackbar is dismissed using "copy"...

projects/core/src/lib/components/toolbar/clicked-coordinates/clicked-coordinates.component.ts
geertplaisier: I meant only the clipboard action, so `this.clipboard.copy(`${evt.mapCoordinates[0].toFixed(decimals)}, ${evt.mapCoordinates[1].toFixed(decimals)}`)` but keep the snackbar message as is. This way people can just paste the numbers in other apps

mprins: +1

geertplaisier (approved):

PR #73, opened 2022-04-01
HTM-273: Prevent application from breaking on unknown WKT

geertplaisier:


mprins (approved):
I will merge this together with a fix in the tailormap-api that linearises any highlight geometry

see also: https://github.com/B3Partners/tailormap-api/pull/85

PR #80, opened 2022-04-21
HTM-156: Implements background layer toggle component

geertplaisier:


matthijsln (changes requested):
Toggle starts to overlap clicked coordinates tool button on smaller screen widths (depends on title of layer of course).

geertplaisier:
I think we should tackle mobile layout as a whole, not per component. The place on the screen is according to design, on mobile we might want to place it somewhere else entirely

matthijsln:
> I think we should tackle mobile layout as a whole, not per component. The place on the screen is according to design, on mobile we might want to place it somewhere else entirely

OK, I see responsive design is not a per-story DoD item.

matthijsln (approved):

PR #81, opened 2022-04-22
HTM-318: Implements live mouse coordinates control

geertplaisier:
Also fixes HTM-326


matthijsln (changes requested):
- The | moves around with the mouse because not every digit font character is the same width. Maybe use two divs where one is justified left and other justified right instead of using a '|' character to separate the numbers.
- The coordinates will be hard to read on a dark map, should have a background.
- Nice that you moved the rounding of coordinates from clicked-coordinates to a general function. Although for degrees maybe more than 4 decimals are needed, looking at [this](https://gis.stackexchange.com/a/8674) I suggest at least 6.

geertplaisier:
Good points, will have a look if we prevent the jumping | 

Regarding layout, this is according to the design. We should specify how it should look with background (colors etc)

We can adjust the rounding, this is code from @mprins, not sure if he had reasons for this rounding, if not we can change it to 6

matthijsln (approved):

PR #87, opened 2022-05-09
Toggle panel content instead of opening/closing on top

geertplaisier:
Implements the 'toggle' behaviour for the side panel on the left


mprins (approved):
LTGM

PR #88, opened 2022-05-13
HTM-319: Drawing UI + interaction

geertplaisier:


mprins:
I don't think this covers all of HTM-319 yet?..

geertplaisier:
> I don't think this covers all of [HTM-319](https://b3partners.atlassian.net/browse/HTM-319) yet?..

Probably true, but the drawing stories are kind of split up, so for me it's not clear what has to be done for what story. The other story this sprint [HTM-311](https://b3partners.atlassian.net/browse/HTM-311) is for example to implement configuration options for point shapes. That is also the reason the text-button does not really do anything yet, this will be implemented in story [HTM-315](https://b3partners.atlassian.net/browse/HTM-315)

mprins (approved):

PR #98, opened 2022-06-09
HTM-314: Add drawing buttons for square, rectangle, ellipse and star

geertplaisier:


mprins (changes requested):
please change "Transparency" to "Opacity" or make the percentage of the slider count the other way around... now transparency of 100% results in an opaque element
image

apart from that things seem to work

PR #117, opened 2022-07-07
Add PDF print export

matthijsln:
HTM-81


projects/core/src/lib/components/print/print/print.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      finalize(() => this.busy$.next(false)),
+    ).subscribe();
+  }
+
+  public downloadMapImage(): void {
+    this.wrapAction(this.mapService.createImageExport(173.4, 130, 150).pipe(
+      tap(dataURL => {
+        const a = document.createElement('a');
+        a.href = dataURL;
+        a.download = 'map.png';
```
mprins, line 79: perhaps it is useful to add the date/time in this filename as well, like the PDF filename

mprins (commented):
The "sqlserver:wegdeel" layer seems to be missing from the image/pdf depending on the zoomlevel of the map, this is with `window.devicePixelRatio` of 1.25 (which is an unzoomed window on my system)

Perhaps it's an idea to warn users about this if they have a divergent DPI?

matthijsln:
> The "sqlserver:wegdeel" layer seems to be missing from the image/pdf depending on the zoomlevel of the map, this is with `window.devicePixelRatio` of 1.25 (which is an unzoomed window on my system)
> 
> Perhaps it's an idea to warn users about this if they have a divergent DPI?

This is indeed a consequence of the 'map scale' differing because the 'DPI' of the web map is different from the DPI of the print map image. Usually the DPI for printing would be higher, so for most situations the WMS server would chose a style that would render more details than are visible on the web map (as a consequence these would be harder to read, smaller labels/lines etc.). But that is arguably less worse than having less details visible on a low DPI which is currently hardcoded in this PR.

I'm going to check out making a new ol.Map instance with a pixelRatio setting, maybe that can solve multiple issues including this one at once.

mprins (approved):
nice!

PR #121, opened 2022-07-14
 HTM-355: Load, highlight and zoom to feature after selection in attribute-list 

geertplaisier:
HTM-412: Fetch feature and highlight on the map after clicking row
HTM-416: Zoom to feature

### DIFF
--- a/projects/map/src/lib/openlayers-map/openlayers-map.ts
+++ b/projects/map/src/lib/openlayers-map/openlayers-map.ts
@@ -15,7 +15,7 @@ import { OpenLayersToolManager } from './open-layers-tool-manager';
 import { OpenLayersEventManager } from './open-layers-event-manager';
 import Feature from 'ol/Feature';
 import Geometry from 'ol/geom/Geometry';
-import { buffer, getCenter } from 'ol/extent';
+import { buffer } from 'ol/extent';
 
 export class OpenLayersMap implements MapViewerModel {
 


projects/map/src/lib/openlayers-map/openlayers-map.ts
```diff
@@ -139,15 +139,8 @@ export class OpenLayersMap implements MapViewerModel {
...
-        .pipe(take(1))
-        .subscribe(mapExtent => {
-          if (containsExtent(mapExtent, geomExtent)) {
-            return;
-          }
-          const center = getCenter(geomExtent);
-          this.zoomTo(center[0], center[1]);
-        });
+      const center = getCenter(geomExtent);
+      this.zoomTo(center[0], center[1]);
```
mprins, line 143: this doesn't zoom the map to the feature, it only moves the feature into the middle of the view; the zoomLevel stays the same, so small features are practically invisible when selected in the list.

I think the existing `OpenLayersMap#zoomToFeature(olFeature: Feature<Geometry>)` should be fixed to call [View#fit(...)](https://openlayers.org/en/latest/apidoc/module-ol_View-View.html#fit)

Unfortunately the story is unclear how much the view should zoom in... but you could:
- get the extent of the feature
- buffer that with n*times the largest axis of the extent
- use that to fit the view

for point features and very small features the buffer would probably need to be larger of you would need a minimum threshold for the extent

mprins (changes requested):
I think zoom-in is not implemented yet and there is lint

projects/map/src/lib/openlayers-map/openlayers-map.ts
mprins: OK; 10 will work for any map projection that uses meter as the UOM, but not for eg. WGS84/degrees, so we will have to revisit

mprins (approved):

PR #124, opened 2022-07-18
Map image export: add scalebar

matthijsln:
HTM-86
HTM-87


mprins (approved):
scalebars exist.

Maybe spend 5 minutes on the width of the DPI input?
image

PR #128, opened 2022-08-10
Add orientation, paper size and image size form controls

matthijsln:
HTM-83
HTM-84
HTM-90


projects/core/src/lib/components/print/print/print.component.css
```diff
@@ -2,3 +2,16 @@
   display: block;
   padding: 14px 14px 0 14px;
 }
+
+::ng-deep .lowerHeight .mat-button-toggle-label-content {
```
geertplaisier, line 6: In case ::ng-deep is needed, we should prefix the selector with something component specific. In this case I think it could easily be changed to:

.lowerHeight ::ng-deep .mat-button-toggle-label-content

This way the CSS is still scoped to the component.

projects/core/src/lib/components/print/print/print.component.html
```diff
@@ -1,17 +1,62 @@
...
+          </mat-form-field>
+          <mat-form-field appearance="fill">
+            <mat-label i18n>Height</mat-label>
+            <input matInput formControlName="height" type="number" min="50" max="600">
+          </mat-form-field>
+          <mat-form-field appearance="fill">
+            <mat-label i18n>DPI</mat-label>
+            <input matInput formControlName="dpi" type="number" min="90" max="600">
+          </mat-form-field>
+          <div class="padding">
```
geertplaisier, line 29: i18n tag missing

projects/core/src/lib/services/map-pdf/map-pdf.service.ts
```diff
@@ -83,11 +83,11 @@ export class MapPdfService {
   private addDateTime(doc: jsPDF, width: number, height: number) {
     const text = $localize `Gemaakt op `;
```
geertplaisier, line 84: Default language is English

projects/core/src/lib/services/map-pdf/map-pdf.service.ts
```diff
@@ -83,11 +83,11 @@ export class MapPdfService {
...
     const date = text + new Intl.DateTimeFormat('nl-NL', { dateStyle: 'full', timeStyle: 'medium'}).format(new Date());
```
geertplaisier, line 85: Instead of using hard-coded nl-NL, it would be better to use the LOCALE_ID injection token to get the current locale so this is also changed when changing the locale in an application:

`@Inject(LOCALE_ID) public locale: string,`

geertplaisier (changes requested):

geertplaisier (approved):

PR #133, opened 2022-08-12
Translations setup

geertplaisier:
Contains initial translation to Dutch.

To run the viewer in Dutch use `npm run start-nl` to get a localized version of the app.
Run `npm run build-localized` to build the application for both English and Dutch (will be added in `dist/en-US` and `dist/nl`.


mprins (approved):
feature info popup stille has "Object Information" in NL, but there are probably others missing as wel as you said

PR #136, opened 2022-08-16
Use the view resolution to set the feature info distance parameter

mprins:
resolves HTM-332


projects/core/src/lib/components/feature-info/feature-info.service.ts
```diff
@@ -30,21 +31,24 @@ export class FeatureInfoService {
...
       .pipe(
         take(1),
-        concatMap(([ layers, applicationId ]) => {
+        concatMap(([ layers, applicationId, resolutions ]) => {
           if (!applicationId || layers.length === 0) {
             return of([]);
           }
           const featureRequests$ = layers.map(layer => {
             const layerId = layer.id;
+            console.log('getFeatures$', resolutions.resolution, coordinates);
```
matthijsln, line 44:
```suggestion

```

matthijsln (changes requested):

matthijsln (approved):

PR #142, opened 2022-08-23
[HTM-442] Create a `version.json` file as an asset

mprins:
this can be requested in other applications for monitoring eg. the admin version info page

see https://github.com/B3Partners/tailormap-admin/pull/118

resolves HTM-442

### DIFF
--- a/bin/prebuild.js
+++ b/bin/prebuild.js
@@ -1,0 +1,12 @@
+const chalk = require('chalk');
+const git = require('git-describe');
+const path = require('path');
+const fs = require('fs-extra');
+const {version: appVersion} = require('../package.json')
+
+const info = git.gitDescribeSync({longSemver: true});
+const file = path.resolve(__dirname, '..', 'projects', 'app', 'src', 'version.json');
+
+fs.writeFileSync(file, JSON.stringify({ version: appVersion, buildDate: Date(), gitInfo: info }), {encoding: 'utf-8'});
+
+console.log(chalk.green(`Wrote version info ${appVersion}, ${info.raw} to ${path.relative(path.resolve(__dirname, '..'), file)}`));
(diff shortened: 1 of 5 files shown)


bin/prebuild.js
```diff
@@ -0,0 +1,15 @@
+const chalk = require('chalk');
+const git = require('git-describe');
+const path = require('path');
+const fs = require('fs-extra');
+const { version: appVersion } = require('../package.json')
+
+const info = git.gitDescribeSync({ longSemver: true });
+const file = path.resolve(__dirname, '..', 'projects', 'app', 'src', 'version.json');
```
geertplaisier, line 8: Check if we could not just add this file to the root of the folder instead of inside projects/app/src. In the angular.json file you specify the path for this asset, I don't think it needs to be inside projects.

bin/prebuild.js
```diff
@@ -0,0 +1,15 @@
...
+
+fs.writeFileSync(
+  file,
+  `{ "version": "${appVersion}", "gitInfo": ${JSON.stringify(info)}}`,
```
geertplaisier, line 12: I think `JSON.stringify({ version: appVersion, gitInfo: info })` would be prettier here.

package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve app",
+    "prebuild": "node bin/prebuild.js",
```
geertplaisier, line 7: When do we run this command?

geertplaisier (changes requested):

package.json
mprins: It runs automatically when you run eg. `npm run build`

mprins: The should be a message in the console

bin/prebuild.js
mprins: > __Warning__
> that is not allowed; 
`An unhandled exception occurred: The projects/app/version.json asset path must start with the project source root.
`
or
`Error: The version.json asset path must start with the project source root.`

PR #144, opened 2022-08-24
Add legend to PDF

matthijsln:
HTM-88


projects/core/src/lib/components/legend/services/legend.service.ts
```diff
@@ -63,6 +63,69 @@ export class LegendService {
...
+          );
+        }));
+      }),
+      map(array => {
+        return array.filter(a => a !== null) as Array<{ appLayer: AppLayerWithServiceModel; imageData: string; width: number; height: number }>;
+      }),
+    );
+  }
+
+  public static imageUrlToPng$(imageUrl: string): Observable<{ imageData: string; width: number; height: number }> {
```
geertplaisier, line 96: Better to place this method in a helper file somewhere since its quite generic

projects/core/src/lib/components/legend/services/legend.service.ts
```diff
@@ -63,6 +63,69 @@ export class LegendService {
...
+          const url = new URL(appLayerWithLegendUrl.url);
+          if (urlCallback) {
+            urlCallback(appLayerWithLegendUrl.layer, url);
+          }
+          return LegendService.imageUrlToPng$(url.toString()).pipe(
+            catchError((e) => {
+              console.log(`Error getting legend from URL ${appLayerWithLegendUrl.url}`, e);
+              return of(null);
+            }),
+            map(legendImage => ({
```
geertplaisier, line 83: If I read correctly this legendImage can be `null`. Not sure what

```{ ...null, appLayer: ... }``` will do but maybe its good to check if legendImage is not null. If its null just return null here, you filter this later.

File: projects/core/src/lib/components/legend/services/legend.service.ts [hunk 3]
```diff
@@ -63,6 +63,69 @@ export class LegendService {
...
+                  ...legendImage,
+                  appLayer: appLayerWithLegendUrl.layer,
+            })),
+          );
+        }));
+      }),
+      map(array => {
+        return array.filter(a => a !== null) as Array<{ appLayer: AppLayerWithServiceModel; imageData: string; width: number; height: number }>;
```
geertplaisier, line 91: `... as ...` is something to prevent. This will effectively disable typechecking here so if something changes somewhere this is not picked up correctly. I supspect you can just just something like `array.filter(Boolean)` to keep Typescript happy. It can then infere the correct type automatically.

mprins:
nice

mprins (approved):

PR #148, opened 2022-08-30
Improve GeoServer legend quality using LEGEND_OPTIONS parameter

matthijsln:
HTM-450

### DIFF
--- a/projects/api/src/lib/models/app-layer-with-service.model.ts
+++ b/projects/api/src/lib/models/app-layer-with-service.model.ts
@@ -1,0 +1,6 @@
+import { AppLayerModel } from './app-layer.model';
+import { ServiceModel } from './service.model';
+
+export interface AppLayerWithServiceModel extends AppLayerModel {
+  service?: ServiceModel;
+}
--- a/projects/api/src/lib/models/app-layer-with-service.model.ts
+++ b/projects/api/src/lib/models/app-layer-with-service.model.ts
@@ -1,6 +1,0 @@
-import { AppLayerModel } from './app-layer.model';
-import { ServiceModel } from './service.model';
-
-export interface AppLayerWithServiceModel extends AppLayerModel {
-  service?: ServiceModel;
-}
--- a/projects/core/src/lib/components/legend/legend-layer/legend-layer.component.ts
+++ b/projects/core/src/lib/components/legend/legend-layer/legend-layer.component.ts
@@ -1,6 +1,7 @@
 import { ChangeDetectionStrategy, Component, Input, OnChanges } from '@angular/core';
-import { AppLayerWithServiceModel, ResolvedServerType } from '@tailormap-viewer/api';
+import { ResolvedServerType } from '@tailormap-viewer/api';
 import { GeoServerLegendOptions, LegendService } from '../services/legend.service';
+import { AppLayerWithServiceModel } from '../../../map/models';
 
 @Component({
   selector: 'tm-legend-layer',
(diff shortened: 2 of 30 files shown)


projects/core/src/lib/components/legend/legend-layer/legend-layer.component.ts
```diff
@@ -7,14 +7,44 @@ import { AppLayerModel } from '@tailormap-viewer/api';
...
   public failedToLoadMessage = $localize `Failed to load legend for`;
 
+  public ngOnChanges() {
+
+    this.urlWithOptions = this.url;
+    this.srcset = '';
+
+    if (this.url) {
+      const url = new URL(this.url);
+      if (url.pathname.includes('/geoserver/') && url.searchParams.get('REQUEST') === 'GetLegendGraphic') {
```
mprins, line 31: Why does this not use the [`ServerTypeHelper`](https://github.com/B3Partners/tailormap-viewer/blob/main/projects/map/src/lib/helpers/server-type.helper.ts) ?

If at some point we start communicating the server type from the backend this is likely to be left out/ignored

matthijsln: Good catch, the server type is already sent but only in the `ServiceModel` but we only have the `AppLayerModel` from the parent component atm.

See other WIP where I do use it: https://github.com/B3Partners/tailormap-viewer/pull/144/files#diff-3fac1d7cb85a7f44a62027cc4c5f9c22aff9cc83d70c050a844cd4352c7b4f7eR116

projects/core/src/lib/components/legend/legend-layer/legend-layer.component.ts
```diff
@@ -1,20 +1,59 @@
...
+  public ngOnChanges() {
+
+    this.urlWithOptions = this.url;
+    this.srcset = '';
+
+    if (this.url) {
+      try {
+        const url = new URL(this.url);
+        if (this.service
+          && (this.service.hiDpiMode === 'geoserver' || (this.service.hiDpiMode === 'auto' && ServerTypeHelper.getFromUrl(this.service.url) === 'geoserver')
```
matthijsln, line 37: This expression is used multiple times in different places, encapsulate in function `ServerTypeHelper.isHiDpiMode(ServiceModel, ServiceHiDpiMode)`

mprins (approved):

projects/api/src/lib/models/app-layer-with-service.model.ts
```diff
@@ -0,0 +1,6 @@
+import { AppLayerModel } from './app-layer.model';
+import { ServiceModel } from './service.model';
+
+export interface AppLayerWithServiceModel extends AppLayerModel {
```
geertplaisier, line 4: Just a suggestion: maybe its good to move this file to `core/src/lib/map/models`. So far the api folder contains mainly models returned by the backend api. This is model is "created" by selectors in map folder so might make more sense there.

projects/map/src/lib/helpers/server-type.helper.ts
```diff
@@ -1,15 +1,36 @@
+import { AppLayerWithServiceModel, ServiceHiDpiMode } from '@tailormap-viewer/api';
 
 export class ServerTypeHelper {
-  public static getFromUrl(url: string): undefined | 'geoserver' | 'mapserver' {
+  public static getFromUrl(url: string): ServiceHiDpiMode.GEOSERVER | ServiceHiDpiMode.MAPSERVER | undefined {
```
geertplaisier, line 4: Maybe its good to rename the ServiceHiDpiMode to a more generic ServerType since its now used in different contexts.

geertplaisier: Now that I'm thinking about it, maybe we could even add this property to the service inside the state directly when retreiving services from the backend, since this won't change and that way its already a property for a service.

PR #151, opened 2022-09-01
[HTM-451][HTM-452] use as-is sorting for the feature info rendering

mprins:
instead of alphabetically sorting the the attributes by name...

The collection of configured attributes is sorted according to the `list_index`, see:

https://github.com/B3Partners/tailormap-persistence/blob/e1dd76a7ca0b0a1a7d280adf0f2eda1ab5e66720/src/main/java/nl/tailormap/viewer/config/app/ApplicationLayer.java#L89-L90

before:
image

after:
image

resolves HTM-451, HTM-452


mprins:
@geertplaisier it seems a bit of a hack, but it works for me. see also:
- https://github.com/angular/angular/issues/31420
- https://github.com/angular/angular/issues/42490

geertplaisier (approved):

PR #153, opened 2022-09-07
Add SCALE to GetLegendGraphic request and gray out layer title when out of scale

matthijsln:
HTM-372


mprins:
It does not seem to work for the "wegdeel" layer which becomes visible above 1:5000 (using chrome on android)

geertplaisier:
For me, Firefox on desktop, it seems to work just fine. Maybe the getResolution$ observable (which uses the mapmove event) works differently on Android/touch

geertplaisier (approved):

PR #159, opened 2022-09-07
HTM-377: Toggle all layers visibility

geertplaisier:


mprins (commented):
I think the initial command should be "Toggle all layers off" according to # 2 in the story 

> When one or more layers is checked/active in the ToC, you can deactive all layers in the ToC through the switch

matthijsln:
It works, but aren't there good svg icons to replace a button with a lot of text on it?

mprins:
> but aren't there good svg icons to replace a button with a lot of text on it?

You could use `visibility_off` and `visibility` icons from the material symbols
visibility_off_FILL0_wght400_GRAD0_opsz48
visibility_FILL0_wght400_GRAD0_opsz48

[visibility_.zip](https://github.com/B3Partners/tailormap-viewer/files/9524521/visibility_.zip)

matthijsln (approved):
👍

PR #162, opened 2022-09-09
[HTM-92] Show print preview polygon for image export

matthijsln:
- [x] Make map image exporter export image with extent as shown by the preview


mprins (approved):
Nice

PR #166, opened 2022-09-14
HTM-471: Added state for filtering on attributes

geertplaisier:


matthijsln (approved):

projects/core/src/lib/filter/models/filter-condition.enum.ts
```diff
@@ -0,0 +1,20 @@
+export enum FilterConditionEnum {
+  UNIQUE_VALUES_KEY = 'UNIQUE_VALUES',
+  NUMBER_EQUALS_KEY = '=',
+  NUMBER_LARGER_THAN_KEY = '>',
+  NUMBER_SMALLER_THAN_KEY = '<',
+  NUMBER_LARGER_EQUALS_THAN_KEY = '>=',
+  NUMBER_SMALLER_EQUALS_THAN_KEY = '<=',
+  NUMBER_BETWEEN_KEY = '><',
+  STRING_EQUALS_KEY = 'EQUALS',
+  STRING_LIKE_KEY = 'LIKE',
```
mprins, line 10: don't we need an ILIKE key as well?

PR #167, opened 2022-09-15
HTM-463: Convert filters to CQL and add to map request

geertplaisier:


projects/core/src/lib/filter/helpers/cql-filter.helper.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(filters.get(1)).toBe('(attribute LIKE \'%value%\')');
+  });
+
+  test('combine multiple filters into a CQL filter', () => {
+    const filterGroup: FilterGroupModel = {
+      id: '1',
+      layerId: 1,
+      filters: [{
+        id: '1',
+        caseSensitive: false,
```
mprins, line 34: actually LIKE is case sensitive, ILIKE is not, see https://docs.geoserver.org/stable/en/user/filter/ecql_reference.html#predicate

```suggestion
        caseSensitive: true,
```

projects/core/src/lib/filter/helpers/cql-filter.helper.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(filters.get(1)).toBe('((attribute LIKE \'%value%\') AND (attribute2 = true) AND (attribute3 = 2020-01-01) AND (attribute4 IS NOT NULL))');
+  });
+
+  test('should create a CQL filter for a tree of filters', () => {
+    const filterGroups: FilterGroupModel[] = [{
+      id: '1',
+      layerId: 1,
+      filters: [{
+        id: '1',
+        caseSensitive: false,
```
mprins, line 78:
```suggestion
        caseSensitive: true,
```

projects/core/src/lib/filter/helpers/cql-filter.helper.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+describe('CQLFilterHelper', () => {
+
+  test('should create a basic CQL filter', () => {
+    const filterGroup: FilterGroupModel = {
+      id: '1',
+      layerId: 1,
+      filters: [{
+        id: '1',
+        caseSensitive: false,
```
mprins, line 14:
```suggestion
        caseSensitive: true,
```

projects/map/src/lib/helpers/ol-layer.helper.ts
```diff
@@ -160,4 +179,18 @@ export class OlLayerHelper {
...
+
+  public static getWmsServiceParams(layer: WMSLayerModel): WmsServiceParamsModel {
+    const params: WmsServiceParamsModel = {
+      LAYERS: layer.layers,
+      VERSION: '1.1.1',
+      QUERY_LAYERS: layer.queryLayers,
+      TRANSPARENT: 'TRUE',
+    };
+    if (layer.filter) {
+      params.CQL_FILTER = layer.filter;
```
mprins, line 191: at some point we probably want to make this a function call that takes `layer.filter` and the `ServerType` or `ResolvedServerType` as an argument and returns the vendor specific filter parameter as noted in https://b3partners.atlassian.net/browse/HTM-463?focusedCommentId=39331

mprins (changes requested):
As noted `LIKE` is cases sensitive, see also https://github.com/B3Partners/tailormap-viewer/pull/166#discussion_r971642819

And I as it is now I think this may rücksichtslos append a `CQL_FILTER`  query parameter, which may of may not cause errors on any other server than Geoserver

mprins (approved):

PR #169, opened 2022-09-21
HTM-462|HTM-461: Added attribute list UI for filtering + implement filtering

geertplaisier:
resolves  HTM-461, HTM-462


mprins (approved):
Fixed the failing test and rebase/squashed some commits for better history

PR #172, opened 2022-09-27
Combine pull and up in 1 command

mprins:
Combine pull and up in 1 command, don't do it quietly to prevent ssh disconnects om main branch


mprins:
this was successful in https://github.com/B3Partners/tailormap-viewer/actions/runs/3134910243/jobs/5089993333

PR #180, opened 2022-10-04
[security] Don't run container as root, add Trivy scanner to scan built images

mprins:


mprins:
see also https://github.com/B3Partners/tailormap-admin/pull/136

PR #181, opened 2022-10-04
[HTM-480] Add hidpi settings and services

matthijsln:


docker/db/docker-entrypoint-initdb.d/3-dump.sql.script
```diff
@@ -48,7 +52,7 @@ ALTER TABLE public.level_ ENABLE TRIGGER ALL;
 ALTER TABLE public.user_ DISABLE TRIGGER ALL;
 
 COPY public.user_ (username, password) FROM stdin;
-admin	{bcrypt}$2a$10$B6My7xzTiTM76Q17.WVvIO8PAsvJuxSGjc0bbUTMCA3kXSuSvycU2
+admin	{bcrypt}$2a$10$yfAqxSZcWXCNSUjTZYvODusTce.B8o8f.pMytcKSPqlJfYwmG0fY6
```
mprins, line 55: this changes the default password no? that means it won't be regenerated in a new deployment

mprins (dismissed):
I don't think the admin password should be changed

docker/db/docker-entrypoint-initdb.d/3-dump.sql.script
mvdstruijk: I tested this config on my local machine and i can login with the password that is generated in the log file. 
So i think this PR is good...

matthijsln: See the comments in https://github.com/B3Partners/tailormap-viewer/blob/HTM-480-update-config-dump-with-hidpi-settings/docker/db/docker-entrypoint-initdb.d/3-load-data.sh

mvdstruijk (approved):

PR #186, opened 2022-10-07
HTM-501: Show 'no results found' message + search status

geertplaisier:
Also fixes HTM-502: Add hint: Type at least 4 characters to start searching


mprins (approved):
🥇

PR #210, opened 2022-11-02
HTM-523: Always register proj4 to OpenLayers

geertplaisier:


projects/map/src/lib/helpers/projections.helper.ts
```diff
@@ -31,4 +28,10 @@ export class ProjectionsHelper {
     return resolutions;
   }
 
+  private static registerProjection(projection: string, definition: string) {
+    if (!proj4.defs(projection) && !!definition) {
```
geertplaisier, line 32: Now we check if the definition does not exist yet. By default proj4 already contains projections for example for `EPSG:3857`. Should we override those with our own definitions or should we just keep the original proj4 definition?

mprins: Not sure I understand... OL also has some built-in projections too, like https://github.com/openlayers/openlayers/blob/a58f238a97804e583d0ec1dc8a5b199513c343fc/src/ol/proj/epsg3857.js#L43

mprins (approved):

PR #231, opened 2022-11-23
[HTM-585] Use POST request for large WMS GetMap URL

matthijsln:


projects/map/src/lib/helpers/ol-layer.helper.ts
```diff
@@ -59,12 +64,13 @@ export class OlLayerHelper {
     };
   }
 
-  public static createLayer(layer: LayerModel, projection: Projection, pixelRatio?: number): TileLayer<TileWMS> | ImageLayer<ImageWMS> | TileLayer<XYZ> | TileLayer<WMTS> | null {
+  public static createLayer(layer: LayerModel, projection: Projection, pixelRatio?: number,
```
geertplaisier, line 67: For readability I think it would be better to move all the arguments to a separate line instead of dividing in 2 lines

geertplaisier (approved):

mprins:
does this also work for proxied services? because there is only a GetMapping in the proxy controller: https://github.com/B3Partners/tailormap-api/blob/main/src/main/java/nl/b3p/tailormap/api/controller/GeoServiceProxyController.java#L85

PR #234, opened 2022-11-24
[HTM-543] Bookmark URL support

puckipedia:
Implements bookmark URLs.

(PR description out of date for now)
~~For now, see commit messages for implementation details.
A lot of the implementation details are not permanent (exact URL format, etc) - this entire system is relatively flexible and 
can even be used via external components (though there is no way to avoid bookmark fragment ID collisions, as of now).~~


puckipedia:
An example bookmark URL:

https://snapshot.tailormap.nl/pull-request/234/feature-bookmark/app/default#@111730.73,468908.19,5!1:4mR8sZQ2kg

projects/core/src/lib/bookmark/bookmark.helper.ts
```diff
@@ -0,0 +1,139 @@
...
+    private static serializeBookmark(bookmark: SerializedBookmark): string {
+        if (bookmark.id === 0) {
+            const encoded = Uint8Array.from(bookmark.value);
+            const decoded = new DataView(encoded.buffer);
+
+            const center = [ decoded.getFloat64(0), decoded.getFloat64(8) ];
+            const zoomLevel = decoded.getFloat32(16);
+            const precision = decoded.getUint32(20);
+
+            return `@${center[0].toFixed(precision)},${center[1].toFixed(precision)},${zoomLevel.toFixed()}`;
```
matthijsln, line 82: Maybe we can spare a few decimals for the zoomLevel

projects/map/src/lib/map/map.component.ts
```diff
@@ -15,15 +18,51 @@ import { MapService } from '../map-service/map.service';
...
+export class MapComponent implements AfterViewInit, OnDestroy {
+  private static BOOKMARK_DESCRIPTOR: BookmarkFragmentStringDescriptor = { type: 'string', identifier: '' };
+
+  private destroyed = new Subject();
 
   constructor(
     private el: ElementRef,
     private mapService: MapService,
-  ) { }
+    private bookmarkService: BookmarkService,
```
geertplaisier, line 29: It would be better to move this code to the `core` project so that the `map` project is not dependant on bookmarks. It would be better to use a service / component to listen to bookmarks and call the map-service API to zoom to locations / update the bookmark.

projects/map/src/lib/openlayers-map/openlayers-map.ts
```diff
@@ -31,6 +31,7 @@ export class OpenLayersMap implements MapViewerModel {
 
   private readonly resizeObserver: ResizeObserver;
   private initialExtent: OpenlayersExtent = [];
+  private initialCenterZoom?: [number[], number] = undefined;
```
geertplaisier, line 34: I don't think this is needed. This is used when app config changes but then you probably don't want to zoom to the same location again.

projects/map/src/lib/map/map.component.ts
```diff
@@ -15,15 +18,51 @@ import { MapService } from '../map-service/map.service';
...
+        }
+
+        const precision = MapSizeHelper.getCoordinatePrecision(measure);
+        this.bookmarkService.updateFragment(MapComponent.BOOKMARK_DESCRIPTOR, `${info.center[0].toFixed(precision)},${info.center[1].toFixed(precision)},${info.zoomLevel.toFixed(1)}`);
+      });
+
+    this.bookmarkService.registerFragment(MapComponent.BOOKMARK_DESCRIPTOR)
+      .pipe(takeUntil(this.destroyed))
+      .subscribe((descriptor) => {
+        if (descriptor === undefined) {
```
geertplaisier, line 45: Isn't there a loop here, when bookmark changes, the map is centered, but on changing the map the bookmark is updated again?

geertplaisier (changes requested):

projects/shared/src/lib/services/bookmark.service.ts
```diff
@@ -0,0 +1,270 @@
...
+    if (data[data.length - 1] !== 0) {
+      return undefined;
+    }
+
+    return new Uint8Array(data.buffer, 0, data.length - 1);
+  }
+
+  private pendingFragments: Map<string, string> = new Map();
+  private fragments: Map<BookmarkFragmentProtoDescriptor | BookmarkFragmentStringDescriptor, BookmarkFragmentInfo> = new Map();
+  private joinedBookmark: Subject<string | undefined> = new Subject();
```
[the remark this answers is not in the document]
puckipedia: If this is a `BehaviorSubject` the URI fragment would immediately be emptied upon loading, which means the application would need logic to prevent that; hence being a plain `Subject`.

projects/shared/src/lib/services/bookmark.service.ts
```diff
@@ -0,0 +1,270 @@
...
+          return existingFragment.observable;
+      }
+
+      let info: BookmarkFragmentInfo | undefined;
+      if (descriptor.type === 'string') {
+          info = {
+            observable: new BehaviorSubject('') as any as BehaviorSubject<any>,
+            serializer: (x: any) => x as string,
+            deserializer: (x: string) => x as any,
+          } as BookmarkFragmentInfo;
```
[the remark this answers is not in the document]
puckipedia: I'm aware; the issue is I kind of painted myself in a corner typewise, which makes these `as any` casts necessary.

projects/shared/src/lib/services/bookmark.service.ts
```diff
@@ -0,0 +1,270 @@
...
+  public registerFragment(descriptor: AnyDescriptor): Observable<any> {
+      const existingFragment = this.fragments.get(descriptor);
+      if (existingFragment !== undefined) {
+          return existingFragment.observable;
+      }
+
+      let info: BookmarkFragmentInfo | undefined;
+      if (descriptor.type === 'string') {
+          info = {
+            observable: new BehaviorSubject('') as any as BehaviorSubject<any>,
```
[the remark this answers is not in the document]
puckipedia: It expects a `BehaviorSubject<any>`, but for TypeScript reasons a `BehaviorSubject<string>` is somehow incompatible.

projects/map/src/lib/openlayers-map/openlayers-map.ts
puckipedia: The issue is that the bookmark may be applied before the map loads and checks the initial extent, which I believe may be able to happen.

projects/map/src/lib/map/map.component.ts
puckipedia: No, the bookmark code will not notify the bookmark has changed, as it hasn't changed noticably.

projects/map/src/lib/map/map.component.spec.ts
```diff
@@ -10,7 +10,7 @@ describe('MapComponent', () => {
       componentProviders: [
         {
           provide: MapService,
-          useValue: { render: mockRender },
+          useValue: { render: mockRender, getMapViewDetails$: jest.fn(), getUnitsOfMeasure$: jest.fn() },
```
geertplaisier, line 13: Not needed anymore

projects/shared/package.json
```diff
@@ -12,6 +12,7 @@
     "rxjs": "^7.5.5"
   },
   "dependencies": {
+    "@bufbuild/protobuf": "^0.4.0",
```
geertplaisier, line 15: Not needed anymore

projects/map/src/lib/map/map.component.ts
geertplaisier: There is still some "loop" here, not infinite but it goes back and forth 1 or 2 times. You notice this when zooming in a little bit, the map starts to zoom but bounces back en zooms in further because of rounding.

geertplaisier (approved):

matthijsln (approved):

PR #244, opened 2022-12-09
HTM-603: Add layer details panel below layers in TOC

geertplaisier:


matthijsln (approved):
Looks cool, layer in panel could follow the selected layer in tree view (why have (i) button at all? looks a bit messy) but UI can always be improved later

PR #258, opened 2022-12-21
HTM-622: Prevent many OR conditions by using GeometryCollection

geertplaisier:


projects/core/src/lib/filter/helpers/cql-spatial-filter.helper.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  public static convertSpatialFilterToQuery(filter: SpatialFilterModel, layerId: number): string | null {
+    if (filter.geometries.length === 0 || filter.geometryColumns.length === 0) {
+      return null;
+    }
+
+    const { baseGeometries, circles } = this.categorizeGeometries(filter.geometries.map(g => g.geometry));
+    const filterGeometries: string[] = [];
+    if (baseGeometries.length > 0) {
+      const geometryCollection = 'GEOMETRYCOLLECTION(' + baseGeometries.join(',') + ')';
```
matthijsln, line 13: We can add case for `length === 1` for the geometry not wrapped in a geometry collection.

matthijsln (changes requested):

matthijsln (approved):

PR #262, opened 2022-12-22
[HTM-546] Add geolocation button

puckipedia:
Unsure how to deal with the coordinate system differences entirely, so using a small `WKTHelper` that reprojects the WKT. This can probably be done nicer.

On press of the button, requests GPS location (continually), zooms to it at first fix (does not follow), and draws an icon plus a "shadow" to indicate the accuracy. On another press, disables showing the current location on the map.

TODO:
  - [x] Circle radius is in degrees with EPSG:4326, so this doesn't work properly with any non-meter units.
  - [x] Show toast on permission denied/error


projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  public ngOnDestroy(): void {
+    this.destroyed.next(null);
+    this.destroyed.complete();
+  }
+
+  private positionSuccess(pos: GeolocationPosition) {
+    this.mapService.getProjectionCode$()
+      .pipe(take(1))
+      .subscribe(projectionCode => {
+          const point = WKTHelper.transformWKT(`POINT(${pos.coords.longitude} ${pos.coords.latitude})`, 'EPSG:4326', projectionCode);
```
geertplaisier, line 64: It feels a bit strange to create WKT, then transform that to WKT in a different projection. Maybe its better to add a method to `projection.helper.ts` to transform coordinates from one projection to another and use that to create WKT in the right projection immediately. Saves a parse/write WKT action. Also prevents the need to do a string-replace to create the CIRCLE WKT

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Store } from '@ngrx/store';
+import { FeatureModel } from '@tailormap-viewer/api';
+
+@Component({
+  selector: 'tm-geolocation',
+  templateUrl: './geolocation.component.html',
+  styleUrls: ['./geolocation.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class GeolocationComponent implements OnInit, OnDestroy {
```
geertplaisier, line 13: Add some unit tests, to check if the expected methods are called when clicking the button.

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.cdr.detectChanges();
+  }
+
+  private positionError(_e: GeolocationPositionError) {
+      this.cancelGeolocation();
+  }
+
+  public onClick(): void {
+    if (this.activeWatch === undefined) {
+      this.activeWatch = navigator.geolocation.watchPosition(this.positionSuccess.bind(this), this.positionError.bind(this), { enableHighAccuracy: true });
```
geertplaisier, line 102: According to the story we should zoom to the location once, show the location and them remove the icon after 10 seconds. Maybe its better to add `watch` later once we have clear how this should work in our case

geertplaisier (changes requested):

mprins (changes requested):
As it is now this does not match the story, either the story should be rewritten or this functionality should be fixed. The story is the authorative source for user documentation.

Also showing the error circle is rather pointless when it is larger than the viewport

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
puckipedia: `ProjectionsHelper` was private, so I decided to instead create a `CoordinateHelper` which is exported, and currently just has a `projectCoordinates` function, which wraps `proj4` (this might be a bit unnecessary, but hey).

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
puckipedia: Added a single unit test to see if `watchLocation` is properly called, and if the map gets zoomed to properly.

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
puckipedia: This should be fixed; I've just made it cap out at 12 seconds (which includes 2 seconds of margin for zooming to the location). The watch stays, as it means that the phone can still correct its location if it gains more accuracy (though the map won't follow after the initial fix)

geertplaisier:
Personally I prefer the current location since it also has to do with 'zooming' and 'moving' the map, but according to the story the button should be next to the coordinate picker. Maybe discuss

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.mapService.getProjectionCode$()
+      .pipe(take(1))
+      .subscribe(projectionCode => {
+          const point = CoordinateHelper.projectCoordinates([ pos.coords.longitude, pos.coords.latitude ], 'EPSG:4326', projectionCode);
+          if (point === undefined) {
+              return;
+          }
+
+          const pointWkt = `POINT(${point[0]} ${point[1]})`;
+          const circleWkt = CoordinateHelper.circleFromWGS84CoordinatesAndRadius([ pos.coords.longitude, pos.coords.latitude ], pos.coords.accuracy, projectionCode);
```
geertplaisier, line 72: Why not 
```
const circleWkt = `CIRCLE(${point[0]} ${point[1]}, ${pos.coords.accuracy})`
```

geertplaisier (changes requested):

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
mvdstruijk: I agree the story is rewritten. This matches the story now

mvdstruijk:
> Personally I prefer the current location since it also has to do with 'zooming' and 'moving' the map, but according to the story the button should be next to the coordinate picker. Maybe discuss

I agree the story is rewritten. This matches the story now

projects/core/src/lib/components/toolbar/geolocation/geolocation.component.ts
puckipedia: the accuracy from the Geolocation API is in meters, but the radius of the OpenLayers(/custom WKT) circle is in degrees. And as a meter -> degree calculation can't be reliably done without its position on the sphere, this uses the same mechanism that OpenLayers uses for its internal geolocation support, which is [a circular polygon](https://openlayers.org/en/latest/apidoc/module-ol_geom_Polygon.html#.circular).

geertplaisier (approved):

PR #263, opened 2022-12-23
 HTM-256: Re-order layers and folders in TOC 

geertplaisier:


mprins (commented):
Does not seem to work on tablet as described in the story, using longpress selects the label text and activates copy widget

Mozilla/5.0 (Linux; Android 12; SM-P610) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36

Browser size 1334 x 632
Screen size 1334 x 800 (Retina)

geertplaisier:
@mprins Dragging/dropping should be fixed now on touch screens

mprins:
Yes, I can drag layers, but now the visibility checkboxes no longer toggle.
Also I can drag group layer into layers and all layers disappear... from the map as well

https://snapshot.tailormap.nl/pull-request/263/dev-htm-256/app/default#@130414.20,458757.97,11.7!1:VWR8UdfuePOnGNZHOg

geertplaisier:
On touch devices or everywhere? Just checked on my iPhone, seems to work alright there

mprins:
I can drag group layer into layers and all layers disappear... from the map as well on desktop
image

toggling layers works on desktop, but not on tablet linked above.

mprins:
https://user-images.githubusercontent.com/1165786/210212339-3d1e8a82-1c2d-41a2-8898-c9b0da7305b9.mp4

puckipedia:
Probably should add a test + protection against dragging a node into its own children.

Also, the drag+drop UX doesn't show which node is being dragged, which I would consider important on mobile (where it's easier to misclick).

<details><summary>If I drag too fast (on Firefox, at least) it seems to leave a "trail" of drop zones.</summary>

image

</details>

geertplaisier:
All issues should be fixed. For re-ordering on touch devices a user first needs to enable re-order mode with a button. This is because scrolling inside the tree would otherwise be impossible.

matthijsln (approved):

PR #269, opened 2023-01-03
map: update vector layers while interacting and animating

puckipedia:
Any features created with `renderFeatures` currently don't update while either animating or interacting. These are used for small overlays such as feature highlighting, and the upcoming geolocation button. For example, when selecting a feature, moving it off screen, then moving back without letting go of the cursor, features are only highlighted once the user lets go of the cursor. This is somewhat inconvenient.


geertplaisier:
Not sure if this is worth it? According to OL docs "but the setting will have a performance impact for large amounts of vector data". Maybe we can make it an internal setting and only apply it to layers where this matters, like GPS position maybe?

puckipedia:
Making the "renders while interacting" flag an option would be fine (though I don't think we're hitting the "large amounts of vector data" any time soon with a few circles and a feature outline :p); though I'd suggest the vector layers should always be updated while animating, for now; if it turns out to be a performance issue later we can always fix it then.

geertplaisier:
Well for the spatial filter we are currently drawing at least 100 features (and maybe more if we increase limits). Not sure what "large amounts" are, I guess 100 should be fine. On the other hand we already have some performance issues, on Firefox in particular, so maybe making it configurable is not that bad?

puckipedia:
ah; i did a quick lookup for `renderFeatures$` and missed that. I'll add a flag then; I suspect 100 might be enough to be a little bit careful with render performance :p

puckipedia:
Okay, it should be a configurable flag now (defaulting to on), as well as disabled for the filter component and  feature info outline.

projects/map/src/lib/map-service/map.service.ts
```diff
@@ -110,9 +110,10 @@ export class MapService {
     featureGeometry$: Observable<FeatureModelType<T> | Array<FeatureModelType<T>>>,
     vectorLayerStyle?: MapStyleModel | ((feature: FeatureModel<T>) => MapStyleModel),
     zoomToFeature?: boolean,
+    updateWhileAnimating: boolean = true,
```
geertplaisier, line 113: Maybe better to just make it optional and only put it to true if you actually want it, which is only for GPS at this moment I think?

projects/map/src/lib/models/vector-layer.model.ts
```diff
@@ -3,4 +3,5 @@ import { LayerTypesEnum } from './layer-types.enum';
 
 export interface VectorLayerModel extends LayerModel {
   layerType: LayerTypesEnum.Vector;
+  updateWhileAnimating: boolean;
```
geertplaisier, line 6: Make it optional?

geertplaisier (changes requested):

geertplaisier (approved):

PR #270, opened 2023-01-03
[HTM-625] [HTM-624] Track initial opacity, add opacity data to bookmark

puckipedia:
TODO:
 - [ ] Check if the backend properly feeds the initial opacity in
 - [ ] Wire "reset opacity" button up properly
 - [ ] Maybe a slightly more compact way to represent opacity (it's currently a float32 in the bookmark, but i'm not sure opacity can be represented as anything but 0-100 currently)
 - [ ] De-DRY / clean up the layer bookmark generation code a bit


projects/core/src/lib/components/toc/toc-node-details/layer-transparency/layer-transparency.component.ts
```diff
@@ -43,10 +47,12 @@ export class LayerTransparencyComponent {
   }
 
   public resetOpacity() {
-    if (!this.layerId) {
-      return;
-    }
-    this.store$.dispatch(setLayerOpacity({ layerId: this.layerId, opacity: undefined }));
-  }
+    this.layer$.pipe(take(1)).subscribe(layer => {
```
geertplaisier, line 50: If you replace `this.layer$` with `this.store$.select(selectLayer(this.layerId))` you can save a public property and initialization.

projects/core/src/lib/map/bookmark/bookmark.helper.ts
```diff
@@ -1,6 +1,11 @@
 import { MapSizeHelper, MapViewDetailsModel, MapUnitEnum } from '@tailormap-viewer/map';
-import { AppLayerModel } from '@tailormap-viewer/api';
 import { TristateBoolean, LayerVisibilityBookmarkFragment, LayerInformation } from './bookmark_pb';
+import { AppLayerWithInitialValuesModel } from '../models';
+
+export interface MapBookmarkContents {
+  visibilityChanges: { id: number; checked: boolean }[];
```
geertplaisier, line 6: Some nitpicking, but maybe rename id: number also to layerId: number to be consisten

geertplaisier (changes requested):

projects/core/src/lib/map/bookmark/bookmark.helper.ts
puckipedia: These are redefinitions of the parameters taken by the `setLayerVisibility` and `setLayerOpacity` actions, which do take `id` and `layerId` respectively. (this is also why it takes `checked` instead of e.g. `visible`.) I considered fixing both but it felt a bit much to refactor that for this PR.

geertplaisier (approved):

PR #286, opened 2023-01-18
HTM-512: Fetch export capabilities and export data from attribute list

geertplaisier:


matthijsln (changes requested):
Let's use https://www.npmjs.com/package/content-disposition-attachment instead of regular expressions...

matthijsln (approved):

PR #295, opened 2023-01-26
Upgrade packages (Angular 15, Material 15, NgRx 15)

geertplaisier:
Upgraded most packages to the latest version. Skipped upgrading jest since some dependencies are not ready.

Please check the layout since a lot has changed in Angular Material 15.


mprins (changes requested):
not looked at code... but
- new "behaviour" for coordinate picker; the toast seems to have an extra outline
image

and the text has a different baseline than the button

matthijsln (approved):

PR #304, opened 2023-02-08
HTM-640: Layer order bookmark

puckipedia:
This stores the list of children of any foreground layer tree node whose contents have been changed from the original.

### DIFF
--- a/projects/core/src/lib/map/helpers/layer-tree-node.helper.ts
+++ b/projects/core/src/lib/map/helpers/layer-tree-node.helper.ts
@@ -13,12 +13,10 @@ export class LayerTreeNodeHelper {
   }
 
   public static getExtendedLayerTreeNode(node: LayerTreeNodeModel): ExtendedLayerTreeNodeModel {
-    if (LayerTreeNodeHelper.isAppLayerNode(node)) {
-      return node;
-    }
     return {
       ...node,
-      expanded: true,
+      expanded: LayerTreeNodeHelper.isAppLayerNode(node) ? undefined : true,
+      initialChildren: node.childrenIds ?? [],
     };
   }
 
@@ -40,7 +38,7 @@ export class LayerTreeNodeHelper {
     };
   }
 
-  public static findLayerTreeNode(layerTreeNodes: LayerTreeNodeModel[], id: string) {
+  public static findLayerTreeNode<T extends LayerTreeNodeModel>(layerTreeNodes: T[], id: string) {
     return layerTreeNodes.find(l => l.id === id);
   }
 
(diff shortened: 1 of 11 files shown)


projects/core/src/lib/map/services/application-map.service.ts
```diff
@@ -100,6 +106,22 @@ export class ApplicationMapService implements OnDestroy {
         }
       });
 
+    combineLatest([
```
geertplaisier, line 109: I think its good to move all bookmark related stuff to a separate service, something like `application-bookmark.service.ts` which you include from `application-map.service.ts` so we can group functionality. The code for bookmarks is already quite long and will probably get longer.

projects/core/src/lib/map/state/map.actions.ts
```diff
@@ -46,6 +46,10 @@ export const moveLayerTreeNode = createAction(
   `${mapActionsPrefix} Move Layer Tree Nodes`,
   props<{ nodeId: string; position: 'before' | 'after' | 'inside'; parentId?: string; sibling?: string; isBaseLayerTree?: boolean }>(),
 );
+export const setLayerTreeNodeChildren = createAction(
+  `${mapActionsPrefix} Set Layer Tree Node Children`,
+  props<{ nodeId: string; children: string[]; isBaseLayerTree?: boolean }>(),
```
geertplaisier, line 51: Why not something like `props<{ nodes: Array<{ nodeId: string; children: string[] }>; isBaseLayerTree?: boolean }>()` so you can send a single update action when reading the bookmark

projects/core/src/lib/map/helpers/layer-tree-node.helper.ts
```diff
@@ -104,7 +102,7 @@ export class LayerTreeNodeHelper {
     layerTreeNodes: ExtendedLayerTreeNodeModel[],
     id: string,
     transformer: (node: ExtendedLayerTreeNodeModel, children: T[]) => T | null): T | null {
-    const node = LayerTreeNodeHelper.findLayerTreeNode(layerTreeNodes, id);
+    const node = LayerTreeNodeHelper.findLayerTreeNode(layerTreeNodes, id) as ExtendedLayerTreeNodeModel;
```
geertplaisier, line 105: Prevent using `as ...` as much as possible, it disables typechecking completely. Here its better to rewrite the `findLayerTreeNode` method to something like:

`public static findLayerTreeNode<T extends LayerTreeNodeModel>(layerTreeNodes: T[], id: string) {`

and now TS will automatically infer that its a `ExtendedLayerTreeNodeModel` since that is what you are passing to the method.

geertplaisier (changes requested):

projects/core/src/lib/map/state/map.actions.ts
puckipedia: There's no clear guidance as to which approach is preferred for actions; I've seen both used. (e.g. `setLayerOpacity`. I adjusted it here, though.

projects/core/src/lib/map/helpers/layer-tree-node.helper.ts
puckipedia: yeah, that was a shortcut on my side, and wasn't sure how `<T extends LayerTreeNodeModel>` would've been accepted. Fixed that now.

projects/core/src/lib/map/services/application-map.service.ts
puckipedia: I kinda underestimated how many bookmark things would actually involve the `ApplicationMapService`, heh.

geertplaisier (approved):

PR #308, opened 2023-02-09
Change default listen address from localhost to 0.0.0.0 in Compose

matthijsln:
Secure by default even when running in host network mode. Requires https://github.com/B3Partners/tailormap-api/pull/336


matthijsln:
Seems like no authenticaton is needed on port 8081, so it definitely shouldn't be exposed.

Maybe we can put the endpoints under port 8080 in a secured manner, so the admin frontend can also use it.

mprins (approved):

PR #317, opened 2023-02-24
Change service, application and appLayer id to strings, add route for service viewer

matthijsln:
Requires https://github.com/B3Partners/tailormap-api/pull/362

Bookmark protobuf generated with prebuild script (needs to be run before starting). Bookmark URLs get kind of long, will add deflating in separate PR.

When creating a geo-service in the admin the ID is now optional instead of always auto-assigned. Decided not to add a form field for service ID, but when POSTing it _can_ be specified. It is more for configuration with manual config files where you want to reference a service by a descriptive service id.


projects/admin-core/src/lib/catalog/geo-service-form/geo-service-form.component.html
```diff
@@ -11,7 +11,6 @@
   <mat-form-field>
     <mat-label i18n>URL</mat-label>
     <input matInput
-           [tmAutoFocus]="!geoService"
```
geertplaisier, line 14: This should be kept. Will autofocus URL field when creating a new service

projects/admin-core/src/lib/catalog/catalog-node-details/catalog-node-details.component.spec.ts
```diff
@@ -91,9 +91,11 @@ describe('CatalogNodeDetailsComponent', () => {
     const { createGeoService$ } = await setup();
     await userEvent.click(await screen.findByText('Add service'));
     expect(await screen.findByText('Create new service')).toBeInTheDocument();
+    await userEvent.type(await screen.findByPlaceholderText('ID'), 'hoi');
```
geertplaisier, line 94: Can be removed

geertplaisier (changes requested):

geertplaisier (approved):

PR #324, opened 2023-03-10
[HTM-545] Google streetview

mprins:
resolves HTM-545


projects/core/src/lib/components/toolbar/streetview/streetview.component.css
```diff
@@ -0,0 +1,3 @@
+.streetview button {
```
geertplaisier, line 1: Is this needed? Should be covered by the `map-control-button-container` class already?

geertplaisier (approved):

projects/core/src/lib/components/toolbar/streetview/streetview.component.css
mprins: indeed, not needed

mprins:
the linting issue is fixed with #331

PR #328, opened 2023-03-14
[HTM-538] F - Location search in ESPG:3857 viewer zooms to EPSG:4326 coordinates

mprins:
- [x] Add a utility to transform a WKT geometry string to a different projection
- [x] reproject location search result to map/view projection

fixes HTM-538

### DIFF
--- a/projects/core/src/lib/components/toolbar/simple-search/simple-search.component.ts
+++ b/projects/core/src/lib/components/toolbar/simple-search/simple-search.component.ts
@@ -1,12 +1,14 @@
-import { Component, OnInit, ChangeDetectionStrategy, inject, OnDestroy } from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject, OnDestroy, OnInit } from '@angular/core';
 import { FormControl } from '@angular/forms';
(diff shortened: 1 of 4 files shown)


projects/core/src/lib/components/toolbar/simple-search/simple-search.component.ts
```diff
@@ -89,14 +86,23 @@ export class SimpleSearchComponent implements OnInit, OnDestroy {
...
       __fid: 'search-result-highlight-feature',
       geometry: searchResult.geometry,
       crs: searchResult.projectionCode,
       attributes: {},
-    }), style, true, true).pipe(
-      takeUntil(timer(5000)),
-    ).subscribe();
+    });
+
+    this.mapService.getProjectionCode$().pipe(take(1))
```
geertplaisier, line 96: This is not how to use RxJS, you should not use a variable outside your stream. There should be a single stream to get the projectioncode and then use switchMap/concatMap to render the features.

geertplaisier (changes requested):

matthijsln (approved):

PR #332, opened 2023-03-17
Combine binary bookmark fragments in a single compressed protobuf

matthijsln:
- [x] Fix weird problem occurring when clicking all checkboxes from top to bottom, when checking BGT other layers change visibility and bookmark goes away, because of `Error decompressing binary bookmark fragment `
- [x] Do not recompress unchanged values everytime bookmark is updated with new location etc
- [ ] ~Check if there is a better compression algorithm, maybe LZMA-JS ?~
- [x] Remove console.log


projects/core/src/lib/map/bookmark/bookmark.proto
```diff
@@ -1,9 +1,23 @@
...
 
 syntax = "proto3";
 package bookmark;
 
+// The bookmark fragments with binary (protobuf) data. Because fragments may use long strings to refer to the same app layer or node ids,
+// these are combined in this message and compressed once instead of compressed separately
+message BinaryBookmarkFragments {
+  // Versioning to allow for backwards compatibility for old bookmarks
+  uint32 version = 1;
+  repeated BookmarkFragment fragments = 3;
```
puckipedia, line 11: Might be useful to use `map<string, bytes> fragments = 2;` here instead? Though if you want consistent bookmarks for consistent state, keeping the list but having them be sorted when put into it would be better..

projects/core/src/lib/bookmark/bookmark.models.ts
```diff
@@ -1,17 +1,79 @@
...
-    };
+export interface BookmarkFragmentDescriptor {
+  get identifier(): BookmarkID;
+  get type(): BookmarkType;
+  get initialValue(): string | any;
+
+  // Return equality of decoded values
+  equals(one: any, other: any): boolean;
+  serialize(value: any): string | Uint8Array;
+  deserialize(value: string | Uint8Array): any;
```
puckipedia, line 14: (Are these `any` because of the type erasure happening later on?)

projects/core/src/lib/bookmark/bookmark.service.ts
```diff
@@ -1,224 +1,210 @@
...
-          if (serialized === undefined) {
-              continue;
-          }
-          if (serialized.length === 0) {
-              continue;
-          }
+    if (binaryFragments.fragments.length > 0) {
+      const protobuf = binaryFragments.toBinary();
+      const deflater = new Deflater({ format: 'raw', level: 9 });
+      deflater.append(protobuf);
```
puckipedia, line 117: Missed this: `deflate.append` may also return blocks, though unlikely at this scale. (This might also explain why the inflate works unreliably - it probably just desyncs the compressed stream.)

projects/core/src/lib/map/bookmark/bookmark.proto
```diff
@@ -1,9 +1,23 @@
...
+// Regenerate bookmark_pb.ts by using `npm run generate-protobuf`
 
 syntax = "proto3";
 package bookmark;
 
+// The bookmark fragments with binary (protobuf) data. Because fragments may use long strings to refer to the same app layer or node ids,
+// these are combined in this message and compressed once instead of compressed separately
+message BinaryBookmarkFragments {
+  // Versioning to allow for backwards compatibility for old bookmarks
+  uint32 version = 1;
```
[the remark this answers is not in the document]
matthijsln: I guess the explicit version field (although quite commonly used) can be removed.

projects/core/src/lib/map/bookmark/bookmark.proto
matthijsln: The fragments are added in sort-order of the 'identifier' in the for-loop:

```typescript
    for(const id of [...fragmentById.keys()].sort()) {
       ...
       binaryFragments.fragments.push(new BookmarkFragment({
    }
```
Didn't change this compared to the previous code. The bookmark should ideally be 'stable' in that the same bookmark should always be encoded the same.

projects/core/src/lib/bookmark/bookmark.models.ts
```diff
@@ -1,17 +1,79 @@
...
-    type: 'string';
-    identifier: string;
+export class BookmarkStringFragmentDescriptor implements BookmarkFragmentDescriptor {
+  private readonly _identifier: BookmarkID;
+
+  constructor(identifier: BookmarkID) {
+    this._identifier = identifier;
+  }
+
+  public get identifier(): BookmarkID {
```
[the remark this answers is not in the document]
matthijsln: Not really, just IDE-generated

projects/core/src/lib/bookmark/bookmark.models.ts
matthijsln: Wasn't completely satisfied with this API, it is because `string` fragments are different from `binary` fragments. Will see if the API can be improved.

projects/core/src/lib/bookmark/bookmark.service.ts
matthijsln: Good catch, I think just using the `deflate()` function is better, no need to work with chunks.

PR #339, opened 2023-03-31
Crud actions for applications. Basic settings only.

geertplaisier:


projects/admin-core/src/lib/application/application-form/application-form.component.ts
```diff
@@ -0,0 +1,96 @@
...
+    name: new FormControl('', { nonNullable: true, validators: [Validators.required] }),
+    title: new FormControl(''),
+    adminComments: new FormControl(''),
+    crs: new FormControl(''),
+    initialExtent: new FormControl<BoundsModel>(EMPTY_BOUNDS),
+    maxExtent: new FormControl<BoundsModel>(EMPTY_BOUNDS),
+    authenticatedRequired: new FormControl(false),
+  });
+
+  public projections = [
```
mprins, line 42: should we use https://github.com/B3Partners/tailormap-viewer/blob/main/projects/map/src/lib/models/projection-codes.enum.ts for this?
and maybe also add the extent of the projection as a boundsmodel

projects/admin-core/src/lib/application/application-form/application-form.component.html
```diff
@@ -0,0 +1,28 @@
...
+    <input matInput placeholder="Title" i18n-placeholder formControlName="title" />
+  </mat-form-field>
+  <mat-form-field>
+    <mat-label i18n>Comments</mat-label>
+    <textarea formControlName="adminComments" matInput></textarea>
+  </mat-form-field>
+  <mat-form-field>
+    <mat-label i18n>Projection</mat-label>
+    <mat-select formControlName="crs" placeholder="Projection" i18n-placeholder>
+      <mat-option *ngFor="let projection of projections" [value]="projection.code">
```
mprins, line 22: It would be nice to have the initial/max bounds update (to the extent of the projection) when choosing a projection.

mprins (approved):

projects/admin-core/src/lib/application/application-form/application-form.component.ts
geertplaisier: Good point, for now I'll keep it as is, because in the future we want to support more projections anyway. Currently the enum you mentioned is only used for search, because there we know in what projection the results are.

projects/admin-core/src/lib/application/application-form/application-form.component.html
geertplaisier: Good idea, will add that

PR #346, opened 2023-04-13
HTM-705 / HTM-704: Admin authorization UI

puckipedia:
Pretty complete, but might need a few licks of paint (mostly around being able to deselect all chips):

2023-04-20-083012_367x129_scrot
2023-04-20-083018_434x289_scrot

### DIFF
--- a/projects/admin-api/src/lib/models/authorization-rules.model.ts
+++ b/projects/admin-api/src/lib/models/authorization-rules.model.ts
@@ -8,7 +8,12 @@ export interface AuthorizationRuleGroup {
   decisions: { [name: string]: AuthorizationRuleDecision },
 }
 
+export enum AuthorizationGroups {
+  ANONYMOUS = 'anonymous',
+  APP_AUTHENTICATED = 'app-authenticated',
+};
+
 export const AUTHORIZATION_RULE_ANONYMOUS: AuthorizationRuleGroup = {
-    groupName: 'anonymous',
+    groupName: AuthorizationGroups.ANONYMOUS,
     decisions: { read: AuthorizationRuleDecision.ALLOW },
 };
--- a/projects/admin-api/src/lib/models/authorization-rules.model.ts
+++ b/projects/admin-api/src/lib/models/authorization-rules.model.ts
@@ -1,0 +1,18 @@
+export enum AuthorizationRuleDecision {
+    ALLOW = 'allow',
+    DENY = 'deny',
+}
+
+export interface AuthorizationRule {
+    decision: AuthorizationRuleDecision,
+}
+
+export interface AuthorizationRuleGroup {
+  groupName: string;
+  decisions: { [name: string]: AuthorizationRule },
+}
+
+export const AUTHORIZATION_RULE_ANONYMOUS: AuthorizationRuleGroup = {
+    groupName: 'anonymous',
+    decisions: { read: { decision: AuthorizationRuleDecision.ALLOW } },
+};
--- a/projects/admin-api/src/lib/models/authorization-rules.model.ts
+++ b/projects/admin-api/src/lib/models/authorization-rules.model.ts
@@ -3,16 +3,12 @@ export enum AuthorizationRuleDecision {
     DENY = 'deny',
 }
 
-export interface AuthorizationRule {
-    decision: AuthorizationRuleDecision,
-}
-
 export interface AuthorizationRuleGroup {
   groupName: string;
-  decisions: { [name: string]: AuthorizationRule },
+  decisions: { [name: string]: AuthorizationRuleDecision },
 }
 
 export const AUTHORIZATION_RULE_ANONYMOUS: AuthorizationRuleGroup = {
     groupName: 'anonymous',
-    decisions: { read: { decision: AuthorizationRuleDecision.ALLOW } },
+    decisions: { read: AuthorizationRuleDecision.ALLOW },
 };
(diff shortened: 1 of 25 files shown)


projects/admin-core/src/lib/shared/components/authorization-edit/authorization-edit.component.html
```diff
@@ -0,0 +1,52 @@
+<div>
+  <div>
+    <mat-label i18n>Allow access from:</mat-label>
+    <mat-chip-listbox [(ngModel)]="selectedChip" (change)="changeAuthenticationType($event.value)">
+      <mat-chip-option value="anonymous">Anyone</mat-chip-option>
```
geertplaisier, line 5: Missing i18n attr on lines 5, 6, 7, 13, 28

projects/admin-core/src/lib/shared/components/authorization-edit/authorization-edit.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        .filter(a => a.name !== 'anonymous' && a.name !== 'app-authenticated')
+        .map(a => ({ name: a.name, used: this.value.find(b => b.groupName === a.name) !== undefined }));
+      this.canAddRule = this.groupList.find(a => !a.used) !== undefined;
+
+      if (notify) {
+          this._onChange(value);
+      }
+
+      // Needed to make sure the initial load of groups works reliably.
+      this.changeDetectorRef.detectChanges();
```
geertplaisier, line 64: I think this should be moved to the end of this function. Now Angular could miss changes to `this.selectedChip`

projects/admin-api/src/lib/models/authorization-rules.model.ts
```diff
@@ -0,0 +1,18 @@
+export enum AuthorizationRuleDecision {
+    ALLOW = 'allow',
+    DENY = 'deny',
+}
+
+export interface AuthorizationRule {
```
geertplaisier, line 6: I think we could lose this extra interface. Why not use just

```
export interface AuthorizationRuleGroup {
  groupName: string;
  decisions: { [name: string]: AuthorizationRuleDecision },
}
```

geertplaisier (changes requested):

projects/admin-api/src/lib/models/authorization-rules.model.ts
puckipedia: Mostly for future extensibility and not having to run a migration, since this interface is shared between api and admin.

geertplaisier (changes requested):
If I refresh the page, with a service or application selected, and the `Anonymous` chip is selected, if I click `Specific groups` there is no list to select from. I first need to select `Anon` again, then `Specific groups` again and I see a list of groups to select from.

projects/admin-core/src/lib/shared/components/authorization-edit/authorization-edit.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      if (notify) {
+          this._onChange(value);
+      }
+
+      const anonymousGroup = this.value.find(a => a.groupName === 'anonymous');
+      const loggedInGroup = this.value.find(a => a.groupName === 'app-authenticated');
+
+      if (anonymousGroup?.decisions?.['read']?.decision === AuthorizationRuleDecision.ALLOW) {
+          this.selectedChip = 'anonymous';
+      } else if (loggedInGroup?.decisions?.['read']?.decision === AuthorizationRuleDecision.ALLOW) {
```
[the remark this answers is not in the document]
puckipedia: It is (kinda); the `app-authenticated` and `anonymous` (system) groups showing up in this list is a different bug that requires some deeper API changes, which would be a bit weird to fix in this PR (and need other backend changes too)

puckipedia:
The state sync bugs (I assume both the groups list failing to update in time and the "not PATCHing properly" bug are linked somehow) don't happen here, so I would suspect something is racey, since that happened earlier in development. If it doesn't PATCH properly, the save button does activate, right? Since that should imply the form is receiving data from the component, i'm not sure where the issue would lie...

projects/admin-core/src/lib/catalog/geo-service-form/geo-service-form.component.html
```diff
@@ -24,4 +24,6 @@
       </mat-option>
     </mat-select>
   </mat-form-field>
+
+  <tm-admin-authorization-edit class="authorization-edit" formControlName="authorizationRules" [groups]="(groups$ | async) ?? []"></tm-admin-authorization-edit>
```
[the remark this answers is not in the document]
puckipedia: Fixed.

projects/admin-core/src/lib/shared/components/authorization-edit/authorization-edit.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  private updateValue(value: AuthorizationRuleGroup[], notify: boolean) {
+      this.value = value;
+      this.updateGroupList();
+
+      if (notify) {
+          this._onChange(value);
+      }
+
+      const anonymousGroup = this.value.find(a => a.groupName === 'anonymous');
```
geertplaisier, line 59: Good to convert these special values to `enum` so they can be reused and easily adjusted if API changes

geertplaisier (changes requested):

geertplaisier (approved):

PR #351, opened 2023-04-20
Select Catalog item after creating (node, geo service, feature source)

geertplaisier:


matthijsln (changes requested):
Does not expand node when adding item in a new folder (which is collapsed by default)

matthijsln (approved):

PR #353, opened 2023-04-21
[HTM-760] Show only add sub folder icon for root level

matthijsln:


geertplaisier:
Ah, good catch... maybe also hide Rename for root levels, since this is never shown anywhere. In fact maybe its better to not have a menu for root, but only an add folder button.

PR #356, opened 2023-04-25
[HTM-727] Disallow blank username and group name

mprins:
also disallow strange (non-alfa-numeric) characters, only `[a-zA-Z0-9]*` are allowed, the Add/Save button is disabled when the form data is invalid.

resolves HTM-727


mprins:
Do you think we should add this `[a-zA-Z0-9]*` restriction for groupname / username in the API as well?

matthijsln:
> @matthijsln @geertplaisier Do you think we should add this `[a-zA-Z0-9]*` restriction for groups[name] / users[username] in the API as well?

Yes, see also [HTM-727](https://b3partners.atlassian.net/browse/HTM-712)

geertplaisier (approved):

PR #362, opened 2023-05-05
HTM-742: Open application from admin

geertplaisier:


matthijsln (changes requested):
Use https://angular.io/api/common/APP_BASE_HREF

PR #369, opened 2023-05-16
[HTM-768] Add GeoService layer authorization editing

puckipedia:


geertplaisier (changes requested):
In general: maybe its possible to move logic from the edit-authorizations component to a helper file with tests. This way the component does not grow so big with so many responsibilities and its easier to test the different scenarios.

geertplaisier (approved):

PR #376, opened 2023-05-26
Update to Angular 16

geertplaisier:


mprins (changes requested):
still has some errors like
```
#14 133.1 ./projects/core/src/public-api.ts:4:0-22 - Error: Module not found: Error: Can't resolve 'typings.d.ts' in '/app/projects/core/src'
#14 133.1 Did you mean './typings.d.ts'?
#14 133.1 Requests that should resolve in the current directory need to start with './'.
#14 133.1 Requests that start with a name are treated as module requests and resolve within module directories (/app, node_modules).
#14 133.1 If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.
#14 133.1 
#14 133.1 ./projects/map/src/lib/helpers/projections.helper.ts:23:9-19 - Error: export 'defs' (imported as 'proj4') was not found in 'proj4' (possible exports: default)
#14 133.1 
#14 133.1 ./projects/map/src/lib/helpers/projections.helper.ts:24:6-16 - Error: export 'defs' (imported as 'proj4') was not found in 'proj4' (possible exports: default)
#14 133.1 
#14 133.1 ./projects/map/src/public-api.ts:4:0-22 - Error: Module not found: Error: Can't resolve 'typings.d.ts' in '/app/projects/map/src'
#14 133.1 Did you mean './typings.d.ts'?
#14 133.1 Requests that should resolve in the current directory need to start with './'.
#14 133.1 Requests that start with a name are treated as module requests and resolve within module directories (/app, node_modules).
#14 133.1 If changing the source code is not an option there is also a resolve options called 'preferRelative' which tries to resolve these kind of requests in the current directory too.
#14 133.1 
#14 133.1 Error: projects/core/src/lib/filter/attribute-filter/attribute-filter.component.ts:174:7 - error TS2322: Type 'string | null' is not assignable to type 'string'.
#14 133.1   Type 'null' is not assignable to type 'string'.
#14 133.1 
#14 133.1 174       return inputValue.toISODate();
#14 133.1           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#14 133.1 
```

and others pointing to `typings`

mprins (approved):

PR #385, opened 2023-05-31
HTM-769: HTM-771: Show configured title in tree and show/edit geoservice

geertplaisier:
HTM-769: HTM-771: Show configured title in tree and show/edit geoservice
from application layer tree


mprins (approved):
lgtm

PR #389, opened 2023-06-02
[HTM-711] Set document.title to app title when loading viewer

mprins:
resolves HTM-711


projects/core/src/lib/pages/viewer-app/viewer-app.component.ts
```diff
@@ -72,6 +74,14 @@ export class ViewerAppComponent implements OnInit, OnDestroy {
       .subscribe(bookmark => {
           this.router.navigate([], { relativeTo: this.route, fragment: bookmark, replaceUrl: true });
       });
+
+    this.store$.select(selectViewerTitle)
+      .pipe(takeUntil(this.destroyed))
+      .subscribe(title => {
```
geertplaisier, line 80: Maybe we should also reset the title? Something like:

```
.subscribe({
    next: (title) => { this.document.title = title || [DEFAULT_TITLE] },
    complete: () => { this.document.title = [DEFAULT_TITLE] } 
});
```

geertplaisier (changes requested):

projects/core/src/lib/pages/viewer-app/viewer-app.component.ts
mprins: make sense

geertplaisier (approved):

PR #395, opened 2023-06-05
[HTM-804] Make it possible to add a description to an app layer in the admin

mprins:
resolves HTM-804, related https://github.com/B3Partners/tailormap-api/pull/485


projects/admin-core/src/lib/application/application-layer-settings/application-layer-settings.component.html
```diff
@@ -28,4 +28,8 @@ <h2>Settings for {{ node.label }}</h2>
     <mat-label>Attribution</mat-label>
     <input matInput formControlName="attribution" />
   </mat-form-field>
+  <mat-form-field>
+    <mat-label>Description</mat-label>
+    <input matInput formControlName="description" />
```
matthijsln, line 33: Maybe make this a multi-line textarea.

Ideally we would also have a preview of the `htmlify` pipe or some documentation.

matthijsln (changes requested):

geertplaisier (approved):

PR #409, opened 2023-06-15
HTM-810: Use Ngrx state for users/groups + listen to SSE events

geertplaisier:


mprins (approved):
some lint needs to be fixed, apart from that LGTM

PR #430, opened 2023-07-13
[do not merge] ghactions test debugging

puckipedia:
HTM-883


puckipedia:
(guesstimation: something in change detection messes up and doesn't properly run CDR after the object gets updated; culminating in the whole form not showing up until after saving? though, the save button activating should cause a CDR cycle?)

puckipedia:
Issue has been found :)

PR #433, opened 2023-07-19
[HTM-854] Add error message to login form to indicate lack of roles

mprins:
image

resolves HTM-854


geertplaisier:
@mprins This is not how it should be done. Inside the @tailormap-viewer/shared library you should not depend on other projects and on the state. So in this case it would be better to provide the details you need through an @Input property on the login form. 

Also referencing other projects like this `import { selectUserDetails } from '../../../../../core/src/lib/state/core.selectors';` is always wrong, they should be referenced by their name (in this case it would be @tailormap-viewer/core, but again, from shared you should not reference other projects anyway.

matthijsln (approved):

mprins:
using "...Rights" for privileges is not actually correct; proper terms include permissions, privileges, roles, grants,...

Rights refer to entitlements determined by law and custom, while privileges refer to special advantages or opportunities granted by power, authority, or circumstance.

PR #441, opened 2023-08-01
[HTM-890] WebGL support

puckipedia:
This enables (optional) WebGL support on the tile-based layers. If WebGL is disabled, it falls back to normal rendering. Vector layers, as well as ImageLayer-based rendering (used for untiled WMS), still use canvas.

Compared to the canvas renderer, if a WebGL tile is partially overlapped due to only some of the tiles of a higher zoom level being loaded, it's not properly clipped but instead overlaps the two tiles, showing both zoom levels simultaneously for transparent layers. Fixing this would require a small-ish patch to OpenLayers.

resolves HTM-890

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -78,8 +78,8 @@
               "budgets": [
                 {
                   "type": "initial",
-                  "maximumWarning": "2.5mb",
-                  "maximumError": "3mb"
+                  "maximumWarning": "3mb",
+                  "maximumError": "3.5mb"
                 },
                 {
                   "type": "anyComponentStyle",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18168,7 +18168,8 @@
         "geotiff": "^2.0.7",
         "ol-mapbox-style": "^10.1.0",
         "pbf": "3.2.1",
-        "rbush": "^3.0.1"
+        "rbush": "^3.0.1",
+        "spectorjs": "^0.9.27"
       },
       "funding": {
         "type": "opencollective",
@@ -18183,6 +18184,22 @@
         "@mapbox/mapbox-gl-style-spec": "^13.23.1",
         "mapbox-to-css-font": "^2.4.1",
         "ol": "^7.3.0"
+      }
+    },
+    "node_modules/ol-mapbox-style/node_modules/ol": {
+      "version": "7.4.0",
+      "resolved": "https://registry.npmjs.org/ol/-/ol-7.4.0.tgz",
+      "integrity": "sha512-bgBbiah694HhC0jt8ptEFNRXwgO8d6xWH3G97PCg4bmn9Li5nLLbi59oSrvqUI6VPVwonPQF1YcqJymxxyMC6A==",
+      "dependencies": {
+        "earcut": "^2.2.3",
+        "geotiff": "^2.0.7",
+        "ol-mapbox-style": "^10.1.0",
+        "pbf": "3.2.1",
+        "rbush": "^3.0.1"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/openlayers"
       }
     },
     "node_modules/on-finished": {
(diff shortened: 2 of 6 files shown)


geertplaisier:
If this does not work for Vector and WMS, is it worth increasing our bundle by 1 MB?

puckipedia:
It will still render those, just not with WebGL.

geertplaisier:
Yeah I understand, but seeing as the bundle size is increased by 1MB which is also bad for performance I was wondering how much benefit we have. Cannot really estimate how much WMTS/XYZ will be used, compared to WMS.

tschaub:
Is the 1MB bundle size increase accurate?  Is it possible to get more details on what is included.  It would be good to ensure that geotiff.js is not getting included (looks like you aren’t using it).

matthijsln:
> Is the 1MB bundle size increase accurate? Is it possible to get more details on what is included? It would be good to ensure that geotiff.js is not getting included (looks like you aren’t using it).

The bundle size only increased by a few KB from 2.48 MB to 2.51 MB, exceeding the configured limit of 2.5 MB. In this PR the limit is increased by 1 MB so the build does not fail but including webgl was just the last straw ;)

mprins:
@puckipedia's OL patch was merged and released in 7.5.0 (#447)

matthijsln:
Merged OL 7.5.1 update. The rendering artifacts are gone which is great and on my desktop no problems. But on Android the page crashes after some zooming and panning with lots of tile loading.

On an iOS device the page crashes after a while too.

mprins:
Crash problems should be er fixed with OL 8.1.0 https://github.com/openlayers/openlayers/releases/tag/v8.1.0

matthijsln:
> Crash problems should be er fixed with OL 8.1.0 https://github.com/openlayers/openlayers/releases/tag/v8.1.0

This PR still has problems with memory usage even when OL would restore the WebGL context after losing it. Some strategies could be worked out but this would still crash TM especially on mobile devices because of high video memory usage.

tschaub:
A couple quick responses.

> 8\. Instead of a cache size per WebGL tile layer, have a global cache to limit the total video memory used

For other layer/source types, the tile cache is on the source.  For WebGL tile layers, we moved this to the layer - since the layer knows a bit more about what it needs (the texture).  One minor downside of this is that if sources are reused across multiple layers, the cache is duplicated (this not extremely common with raster layers).  If the cache were moved two levels up - to the map renderer - it could be shared across layers and sources.  I think it makes sense to go in this direction.

If you lower the `cacheSize` option for the layer, are you seeing reduced memory use?  I'm curious if what you are seeing looks like a leak, or just high use.


> 13\. Do not render a tile from a lower level when we already know tiles are available from a higher level that will completely draw over that tile.

The tile renderer currently considers child tiles (at z+1) before considering parents (at <z).  See https://github.com/openlayers/openlayers/blob/v8.1.0/src/ol/renderer/webgl/TileLayerBase.js#L584-L635.  So we never should go deeper than z+1.  We could limit the conditions under which we consider child tiles (if the target zoom is > the tile zoom, for example) - we currently pick make this decision based on which direction the user is zooming.  But it may lead to uglier rendering if we always prefer lower z tiles.

Thanks for your efforts on this.  I want to make time to read through your ideas above more thoroughly.  And I'd like to help make WebGL tile rendering work for your use case.

matthijsln:
Thanks for your input!

Looks like the WebGL renderer is already pretty smart about reducing the amount of tiles rendered!

For the shared cache, if the only upside is having a single memory limit, that could still be achieved in the current API by setting a `cacheSize` taking into account the number of layers and the tile size of each layer. So not sure refactoring the cache would be be necessary.

About our use case:

The reason we're looking at using WebGL is that our example with a WMTS base layer and 5 WMS overlays is pretty slow, but only on tilePixelRatio>1 when enabling high-dpi rendering for all layers. Using high-dpi rendering for this many overlays might just not be suitable for either canvas or WebGL rendering, regardless of how much it's optimized. It's just too many pixels to push I think.

Our example is kind of a out-there example when you look at most web gis viewers which use way fewer overlays, but our admin tool allows users to easily configure a viewer lots of WMS layers and we enable hi-dpi rendering for WMS automatically so hitting a low-fps situation is kind of easy.

I think having only hi-dpi for background layers and no hi-dpi rendering for WMS by default and keep using the canvas renderer is the best for now, and to revisit WebGL rendering when adding support for using the existing vector layer features in OpenLayers on our end. This would achieve even better crisp vector lines compared to WMS, obviously ;)

PR #459, opened 2023-08-24
[HTM-838] Add delete feature option for edit component

mprins:


projects/core/src/lib/components/edit/edit-dialog/edit-dialog.component.ts
```diff
@@ -123,6 +123,32 @@ export class EditDialogComponent implements OnInit {
...
+          return this.editFeatureService.deleteFeature$( viewerId, layerId, {
+            __fid: currentFeature.feature.__fid,
+            attributes: currentFeature.feature.attributes,
+          });
+        }),
+      )
+      .subscribe(succes => {
+        if (succes) {
+          this.closeDialog();
+          this.mapService.getLayerManager$()
```
geertplaisier, line 143: Maybe we can add this as 'shortcut' inside the MapService so its easier to use, something like 
`mapService.refreshLayer(layerId)` where the MapService then executes this code. This is probably something we do more often in the future.

projects/api/src/lib/services/tailormap-api-v1.service.ts
```diff
@@ -90,11 +90,17 @@ export class TailormapApiV1Service implements TailormapApiV1ServiceModel {
...
-      `${TailormapApiConstants.BASE_URL}/${params.applicationId}/layer/${params.layerId}/edit/feature/${params.feature.__fid}`,
-    );
-  }
+    public deleteFeature$(params: { applicationId: string; layerId: string; feature: FeatureModel }): Observable<HttpStatusCode> {
+        return this.httpClient.delete<HttpResponse<Response>>(
+            `${TailormapApiConstants.BASE_URL}/${params.applicationId}/layer/${params.layerId}/edit/feature/${params.feature.__fid}`,
+            { observe: 'response' },
+        ).pipe(
+            map(response => {
+                    return response.status as HttpStatusCode;
```
geertplaisier, line 99: `... as HttpStatusCode` is probably an uneeded cast, `.status` should already be this type?

geertplaisier (changes requested):

projects/api/src/lib/services/tailormap-api-v1.service.ts
mprins: `HttpResponseBase.status` is a number

projects/core/src/lib/components/edit/edit-dialog/edit-dialog.component.ts
mprins: ✅

PR #461, opened 2023-08-29
[HTM-662] create new feature functionality

mprins:
start with a button...

TODO
- [x] get layer attribute details for edit form
- [x] create new feature from layerdetails to feed edit form
- [x] show edit dialog
- [x] enable create geometry in native tool with `layerDetails.details.geometryType`
- [x] disable mapclick/feature info call
- [x] save new feature
- [x] ...


projects/core/src/lib/components/edit/helpers/edit-model.helper.ts
```diff
@@ -1,7 +1,6 @@
-import {
-  AttributeModel, AttributeType, ColumnMetadataModel, FeatureModel, LayerDetailsModel,
-} from '@tailormap-viewer/api';
+import { AttributeModel, AttributeType, ColumnMetadataModel, FeatureModel, LayerDetailsModel } from '@tailormap-viewer/api';
 import { FormFieldModel } from '../models/form-field.model';
+import { $localize } from '@angular/localize/init';
```
geertplaisier, line 3: Should be removed. Must only be imported once (in polyfills.ts)

projects/core/src/lib/components/edit/edit/edit.component.html
```diff
@@ -26,4 +26,19 @@
     </ng-container>
   </div>
 
+  <button [matMenuTriggerFor]="geometryTypeMenu"
+          [class.disabled]="this.layer.value === null"
```
geertplaisier, line 30: Not really wrong but using `this` is not needed here and below

geertplaisier (changes requested):

geertplaisier (approved):

PR #494, opened 2023-10-17
[HTM-932] Fix map shift for web mercator projection

matthijsln:
The OpenLayers projection was initialized with the extent set to the max extent the admin configured for the app. This is obviously wrong, but even if the app max extent was set to the EPSG:3857 extent in the admin (or the code would use the projection extent from `/api/app/.../map` member `crs.bounds`), there would be a mismatch for the latitude range (official [EPSG:3857](https://epsg.io/3857) goes from [-85.06, 85.06] but the projection extent for OL should be [-90.0, 90.0].  As a fix just use the string as projection instead of creating a new OL Projection. OL recognizes the 'EPSG:3857' string and uses the correct extent. For other projections such as EPSG:28992 this change does not seem to matter.

Also actually applies the max map view extent.

TODO:
- [x] Resolutions are still calculated from app max extent


matthijsln:
Projection extent does not equate to tile grid extent. Resolutions can't be calculated from projection. For now just do not set `View.resolutions` at all, we don't use fixed resolutions anyway. Services have tile grids.

mprins (approved):

PR #514, opened 2023-11-10
HTM-945: Add making attributes readonly to feature type form

geertplaisier:


matthijsln:
In Dutch I only can read 'Ingeschal' and 'Bewerkbaa'

mprins (approved):

PR #545, opened 2024-01-12
 HTM-973, HTM-678: Expand only folders with checked/visible layers in TOC 

geertplaisier:


matthijsln (approved):

matthijsln:
Also fixes HTM-678 (initial extent is retrieved on load even when bookmark exists with different center)

PR #561, opened 2024-02-08
Do not navigate to same URL after error again to avoid infinite routing loop, fix navigating to empty URL

matthijsln:


matthijsln:
Navigation errors seem to be related to https://github.com/B3Partners/tailormap-viewer/pull/557/commits/73a312c4506145d56ad0d6e35e1501982accae94#diff-99b494df2f01942923375ad443b39827d39ebfc10c53a39e1b39c03403854c37

projects/core/src/lib/services/navigation-error-router.service.ts
```diff
@@ -8,24 +8,30 @@ import { filter } from 'rxjs';
...
 })
 export class NavigationErrorRouterService {
+  private urlNavigatedToAfterError: string | null = null;
+
   constructor(router: Router, @Inject(APP_BASE_HREF) baseHref: string, @Inject(LOCALE_ID) localeId: string) {
     // Use an alternative catch-all route instead of '**', this allows libraries to add routes
     router.events.pipe(
       takeUntilDestroyed(),
       filter((event): event is NavigationError => event instanceof NavigationError),
-    ).subscribe((event: NavigationError) => {
```
geertplaisier, line 16: Why remove type?

projects/core/src/lib/services/navigation-error-router.service.ts
```diff
@@ -8,24 +8,30 @@ import { filter } from 'rxjs';
...
+    ).subscribe((event) => {
       console.error('Navigation error', event);
```
geertplaisier, line 19: Maybe we can log the error at the end of the method? If we are doing a `navigateByUrl` then the console.error might not be needed since the user will end up on the right spot? If the error is triggered again by the same URL (so the `if` statement returns false) we then can show an error?

geertplaisier (changes requested):

geertplaisier (approved):

PR #563, opened 2024-02-12
HTM-950: Show only layers matching application projection

geertplaisier:


matthijsln:
This change unfortunately breaks xyz layers, which don't have projection info.

matthijsln:
> This change unfortunately breaks xyz layers, which don't have projection info.

Fixed in HTM-1009

matthijsln (approved):

PR #570, opened 2024-02-13
HTM-962: Override GeoTools' projection WKT with proj4js definition for EPSG:28992

matthijsln:


matthijsln:
The difference is in the reversed sign for the rx and ry towgs84 parameters. GeoTools reprojection with these parameters works correctly. Not sure where the difference is, let's just keep this workaround for now.

PR #582, opened 2024-02-16
HTM-1025: Application name should be unique

geertplaisier:
It would be good to also check this on the backend / make name a UNIQUE field in the database?


matthijsln:
https://github.com/B3Partners/tailormap-api/pull/723

PR #604, opened 2024-04-05
HTM-1082: Implement configured form in viewer edit component

geertplaisier:


matthijsln (changes requested):
'Allow the user to enter custom input' is not checked, yet the input is still editable but the value changes to 'null' after blurring the input.

Screenshot from 2024-04-05 10-15-29

https://github.com/B3Partners/tailormap-viewer/assets/1926261/86ddc949-181e-4512-bf7d-121da4e49498

PR #613, opened 2024-04-18
Add option for full-screen dialogs

geertplaisier:
Not currently used in the viewer, but can be passed as an option to enable dialogs to be fullscreen


matthijsln:
We can also re-use this icon to enable full screen for the entire window ;)

matthijsln (approved):

PR #620, opened 2024-05-01
HTM-1104: use tm-admin-password-field component for client-secret

mvdstruijk:
Hide OIDC client-secret field


matthijsln (approved):
While this changes the client-side input to a password field, the client secret is still available via the API and logged in admins can see it using F12. Although a OIDC client secret is usually tied to a specific redirect URL, so leakage is not immediately problematic.

In the tailormap-api repository you can add 
```
@JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
```
to the `clientSecret` property just as for the `password` property on the `User` entity.

This won't show in the interface whether a clientSecret was set or not though. We could add a `isClientSecretSet()` and use that if that is desired though (but we don't have that for user passwords either).

PR #633, opened 2024-05-24
HTM-1106: Allow setting to automatically refresh layers

geertplaisier:


projects/core/src/lib/map/services/application-layer-refresh.service.ts
```diff
@@ -0,0 +1,61 @@
...
+          .filter((l): l is RefreshableLayer => typeof l.autoRefreshInSeconds === 'number');
+        this.setRefreshLayers(refreshableLayers);
+      });
+  }
+
+  public setRefreshLayers(refreshableLayer: RefreshableLayer[]) {
+    const ids = new Set(refreshableLayer.map(r => r.id));
+    const currentIds = Array.from(this.refreshingLayers.keys());
+    currentIds.forEach((key) => {
+      if (!ids.has(key)) {
```
matthijsln, line 43: Once turned on, the refresh time can't be changed (like opacity can) unless layer is hidden again, but we can always add that in the future because we don't have a control like opacity at the moment anyway.

matthijsln (approved):

PR #647, opened 2024-06-12
HTM-1125: Allow to copy an entire application

geertplaisier:


matthijsln (changes requested):
Save button is disabled and only enabled after editing the form.

mprins (commented):
nice. 
Apart from the issue that @matthijsln pointed out regarding the save button in the popover form

mprins (approved):

PR #650, opened 2024-06-17
HTM-870: display allowed origins for WMS

matthijsln:
Not sure if actually useful.


matthijsln:
Not that useful after all.

PR #659, opened 2024-06-21
HTM-1093: Add Search Index CRUD pages

geertplaisier:


mprins:
er missen nog wat vertalingen

mprins (changes requested):
I've added the missing language resources, I think the location of the Feature type button would make more sense under the "feature source en type" heading instead of just floating about... 

image

In any case the width and wordwrapping of the button needs to be fixed


image

image

mprins (approved):

PR #712, opened 2024-09-20
HTM-1160: Implementing markdown and simple template for feature info

geertplaisier:
[HTM-1160](https://b3partners.atlassian.net/browse/HTM-1160) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)

[HTM-1160]: https://b3partners.atlassian.net/browse/HTM-1160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


projects/core/src/lib/components/feature-info/feature-info-template-renderer/feature-info-template-renderer.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    from(import('marked'))
+      .pipe(take(1))
+      .subscribe(markedLib => {
+        const replacements: string[] = [];
+        const replacementMap = new Map<string, string>();
+        this.feature?.sortedAttributes.forEach(a => {
+          replacements.push(a.key);
+          replacementMap.set(a.key, a.attributeValue);
+        });
+        const regex = new RegExp('{{\\s*(' + replacements.join('|') + ')\\s*}}', 'gi');
```
matthijsln, line 37: While not a good idea a attribute name could contain regexp characters, should we use [RegExp.escape](https://github.com/es-shims/RegExp.escape)?

projects/core/src/lib/components/feature-info/feature-info-template-renderer/feature-info-template-renderer.component.ts
```diff
@@ -0,0 +1,46 @@
...
+        const replaced = this.template?.replace(regex, (fullMatch, match) => {
+          return replacementMap.get(match) || "";
```
matthijsln, line 39: So attribute values can contain markdown and html tags (with javascript code sanitized). Do we want escape some  of this? Or replace after markdown to html conversion?

projects/core/src/lib/components/feature-info/feature-info-template-renderer/feature-info-template-renderer.component.ts
geertplaisier: Good point, added an esacape method

projects/core/src/lib/components/feature-info/feature-info-template-renderer/feature-info-template-renderer.component.ts
geertplaisier: Think it does not harm to allow markdown in attribute values? Should not be a problem from security perspective since we sanitize last?

matthijsln (approved):

PR #729, opened 2024-10-18
HTM-1264: Make long layer titles visible in TOC

steinkobben:
[HTM-1264](https://b3partners.atlassian.net/browse/HTM-1264) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)

* Set height of tree-node in toc-tree to fit-content.
* Made text in .tree-node__label-content wrap.

[HTM-1264]: https://b3partners.atlassian.net/browse/HTM-1264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/core/src/lib/components/toc/toc-node-layer/toc-node-layer.component.css
+++ b/projects/core/src/lib/components/toc/toc-node-layer/toc-node-layer.component.css
@@ -15,6 +15,10 @@
   align-items: center;
 }
 
+.tree-node__label-content {
+  white-space: wrap;
+}
+
 .tree-node__label.modified {
   font-weight: bold;
 }
--- a/projects/core/src/lib/components/toc/toc-node-layer/toc-node-layer.component.css
+++ b/projects/core/src/lib/components/toc/toc-node-layer/toc-node-layer.component.css
@@ -9,6 +9,10 @@
   line-height: 18px;
 }
 
+.tree-node {
+  padding: 5px 0;
+}
+
 .tree-node__label {
   padding-right: 16px;
   display: flex;
@@ -17,6 +21,7 @@
 
 .tree-node__label-content {
   white-space: wrap;
+  overflow-wrap: break-word;
 }
 
 .tree-node__label.modified {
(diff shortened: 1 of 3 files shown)


projects/core/src/lib/components/toc/toc/toc.component.css
```diff
@@ -8,8 +8,16 @@
   padding-right: var(--toc-side-padding);
 }
 
+.toc-tree ::ng-deep .mat-tree-node {
+  height: fit-content !important;
```
geertplaisier, line 12: Do not use !important unless really needed, in this case I don't think its needed at all

projects/core/src/lib/components/toc/toc/toc.component.css
```diff
@@ -8,8 +8,16 @@
...
+}
+
+.toc-tree ::ng-deep .tree-node-wrapper {
+  height: fit-content !important;
```
geertplaisier, line 16: See above

projects/core/src/lib/components/toc/toc-node-layer/toc-node-layer.component.css
```diff
@@ -15,6 +15,10 @@
   align-items: center;
 }
 
+.tree-node__label-content {
+  white-space: wrap;
```
geertplaisier, line 19: Maybe add `overflow-wrap: break-word;` as well

geertplaisier (commented):
On the .tree-node class we might want to add some padding (`padding: 5px 0;` for example) so there is some room above and below the label if the label is over multiple lines. For that to work we need to remove the `.tree-node__actions ...` styling from `toc-node-layer.component.css` and put `.tree-node__actions { display: none; }` instead

geertplaisier (approved):

PR #738, opened 2024-10-31
HTM-1250: User interface to administer tasks

steinkobben:
[HTM-1250](https://b3partners.atlassian.net/browse/HTM-1250) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)

[HTM-1250]: https://b3partners.atlassian.net/browse/HTM-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mprins:
**Please use "Squash and merge" to get a clean commit history**

geertplaisier (changes requested):

projects/admin-api/src/lib/models/task-details.model.ts
```diff
@@ -0,0 +1,15 @@
...
+  cronExpression: string;
+
+  timezone?: string;
+  startTime?: Date;
+  lastTime?: Date;
+  status?: string;
+  progress?: string;
+  lastResult?: string;
+  nextFireTimes?: Date[];
+  jobData?: Map<string, string>;
```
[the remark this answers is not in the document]
mprins: It's a `HashMap<String, Object>` in the Java API, a collection of key/value pairs, the value is not neccesarily a String (could be a number or even an object as well - depending on the kind of task), example in the API: https://github.com/Tailormap/tailormap-api/blob/ce8a0dca0678a453d3a8b45f5e21b88d99058037/src/main/java/org/tailormap/api/controller/admin/TaskAdminController.java#L181-L195

projects/admin-core/src/lib/templates/admin-template/navigation/navigation.component.ts
```diff
@@ -79,6 +79,13 @@ export class NavigationComponent implements OnInit {
       icon: 'admin_application',
       requireAdmin: true,
     },
+    {
```
geertplaisier, line 82: Maybe better to put this in the bottomButtons instead

geertplaisier (changes requested):

projects/admin-core/src/lib/tasks/task-details/task-details.component.html
```diff
@@ -0,0 +1,89 @@
+<h2 i18n="@@admin-core.tasks.task.information">Task Information</h2>
+
+<ng-template #infoString let-infoType="infoType" let-infoValue="infoValue">
```
geertplaisier, line 3: Not used anymore since we have a component for this now?

projects/admin-core/src/lib/tasks/services/task-monitoring.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  public deleteTask(uuid: string, type: string ) {
+    this.adminApiService.deleteTask$(uuid, type)
+      .pipe(
+        catchError(() => {
+          return of({ error: $localize `:@@admin-core.tasks.error-deleting-task:Error while deleting task` });
+        }),
+        map(response => {
+          if (ApiResponseHelper.isErrorResponse(response)) {
+            return TasksActions.deleteTaskFailed({ error: response.error });
```
geertplaisier, line 72: This won't do anything. Returning an action here does not execute this (would need this.store$.dispatch(...) for that. Also see my comment below about calling the method. Maybe you want to rewrite this logic a bit (you might for example not need to add the failed message to the state at all but just show it in the UI where it is called). For the success you will need to call the state to remove the item from the list.

projects/admin-core/src/lib/tasks/task-details/task-details.component.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (task) {
+          this.confirmDelete.confirm$(
+            $localize `:@@admin-core.tasks.delete-task:Delete task \'${task.description}\'`,
+            $localize `:@@admin-core.tasks.delete-task-message:Are you sure you want to delete task \'${task.description}\'? This action cannot be undone.`,
+            true,
+          )
+            .pipe(
+              take(1),
+              filter(answer => answer),
+              tap(() => this.taskMonitoringService.deleteTask(task.uuid, task.type)),
```
geertplaisier, line 75: It would be better to wait for this one to complete. Now you always redirect to /admin/tasks, even if removing the task was unsuccessful. This means you would need to remove the .subscribe from the taskMonitoringService.deleteTask method and just return the observable to be called here

projects/admin-core/src/lib/templates/admin-template/navigation/navigation.component.ts
geertplaisier: Not solved yet

geertplaisier (changes requested):

PR #787, opened 2024-12-19
HTM-1369: Added municipality filter to simple search component

geertplaisier:
[HTM-1369](https://b3partners.atlassian.net/browse/HTM-1369) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)

[HTM-1369]: https://b3partners.atlassian.net/browse/HTM-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mprins (approved):
added some useful test data in the api setup, see https://github.com/Tailormap/tailormap-api/pull/1094

matthijsln (changes requested):
- Searching for a municipality with spaces does not work, for instance 'Valkenburg aan de Geul'. It seems quoting with single or double quotes doesn't work... Maybe switch to gemeentecode? Getting codes and names from the PDOK WFS or OGC API is very slow as the geometry is always returned, but they can be gleaned from https://api.pdok.nl/kadaster/bestuurlijkegebieden/ogc/v1/api?f=json
- User needs to know the exact spelling of municipality names the API accepts, how a user might spell it might be different from what the API uses. So an autosuggest thing would be helpful (maybe using the JSON above, not sure how 'official' that API is though)

PR #846, opened 2025-03-24
Save correct index for 3D tiles layers

steinkobben:
Bugfix: Save the correct index for 3D Tiles layers when multiple are created at the same time.


projects/map/src/lib/openlayers-map/cesium-map/cesium-manager.ts
```diff
@@ -121,11 +121,16 @@ export class CesiumManager {
           const primitive = scene3d.primitives.get(this.layers3d.get(layer.id) ?? 0);
           primitive.show = true;
         } else {
-          this.layers3d.set(layer.id, scene3d.primitives.length);
           const tiles3DLayerPromise = this.create3dLayer(layer);
           tiles3DLayerPromise.then(tiles3dLayer => {
             if (tiles3dLayer) {
               scene3d.primitives.add(tiles3dLayer);
+              for (let i = 0; i < scene3d.primitives.length; i++) {
```
geertplaisier, line 128: Why not `scene3d.primitives.find(...)` instead of a for loop?

geertplaisier (approved):

PR #898, opened 2025-06-11
HTM-1554: Refresh layer + attribute list after edit feature

geertplaisier:


projects/core/src/lib/components/attribute-list/services/attribute-list-data.service.ts
```diff
@@ -156,4 +150,22 @@ export class AttributeListDataService implements OnDestroy {
...
+    source$
+      .pipe(
+        takeUntil(this.destroyed),
+        withLatestFrom(this.store$.select(selectAttributeListTabs)),
+        map(([ sourceResult, tabs ]) => {
+          return tabs.filter(tab => typeof tab.layerId === 'undefined' ? false : shouldUpdateTab(sourceResult, tab.layerId));
+        }),
+      )
+      .subscribe(tabs => {
+        tabs.forEach(tab => {
```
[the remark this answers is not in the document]
geertplaisier: updatePage will also trigger loading data, now it was loading data twice.

projects/core/src/lib/components/attribute-list/services/attribute-list-data.service.ts
```diff
@@ -156,4 +150,22 @@ export class AttributeListDataService implements OnDestroy {
...
+          this.store$.dispatch(AttributeListActions.setHighlightedFeature({ feature: null }));
```
steinkobben, line 164: Could be outside of forEach loop

geertplaisier: Done

steinkobben (approved):

PR #925, opened 2025-07-03
Show drawing styles from library and drawing improvements

matthijsln:
- [x] Display drawing styles for use (with proper layout, description)
- [x] Display style preview for saved styles (similar to object list), DrawingFeatureIconComponent works on feature not style, must also have `type` 
- [x] Style is not updated in object list after changing for current feature
- [x] Display style preview for IMAGE style
- [x] For IMAGE style only show scale (percent, from ~50-200%)/rotation style controls
- [x] Display IMAGE style features in object list correctly (icon margin not correct yet)
- [x] After drawing saved style then switching to 'Draw point', marker image is still used
- [x] Double click on object list should edit description
- [x] New objects should have default description from saved style or from shape
- [x] MARKER features size incorrect
- [x] Label for IMAGE features not at correct position
- [x] When selecting an IMAGE or POINT object in object list, the type-dependent style editors do not change
- [x] Show selection rectangle for IMAGE or POINT
- [x] Draw STAR style preview
- [x] Update style previews
- [x] Use same SVG paths in map as in UI for cross, star, arrow shapes
- [x] Show selection rectangle for POINT and IMAGE
- [x] Scroll to selected object in object list


mprins (commented):
goed bezig!

geertplaisier (approved):

PR #930, opened 2025-07-11
Handle filters for hidden attributes

steinkobben:
- Use admin api unique values service in admin
- Prevent creating filters on hidden attributes
- Disable and hide filters for hidden attributes in viewer


projects/core/src/lib/components/filter/filter-list-item/filter-list-item.component.ts
```diff
@@ -22,7 +22,19 @@ export class FilterListItemComponent {
...
   public set filterGroup(filterGroup: ExtendedFilterGroupModel | null) {
-    this.filter = filterGroup;
+    if (!filterGroup) {
+      return;
+    }
+    this.filter = {
+      ...filterGroup,
+      filters: filterGroup?.filters.filter(f => {
+        if (FilterTypeHelper.isAttributeFilter(f) && f.attributeNotFound) {
+          console.error(`Filtered attribute '${f.attribute}' not found. Filter hidden and disabled.`);
```
matthijsln, line 32:
```suggestion
          console.error(`Attribute '${f.attribute}' not found. Filter hidden and disabled.`);
```
Attributes are 'hidden', not filtered.. Attributes could also have been removed or renamed.

matthijsln (changes requested):
Disabling and hiding filters only seems to hide it from the UI (sometimes). Filter is still applied in GetMap request and attribute list.

See screenshot below. 

Image

matthijsln:
image

Filter is now hidden and not applied anymore, can still leave empty filter block but that is harmless.

matthijsln (approved):

PR #941, opened 2025-07-22
Enable value change events while sliding

matthijsln:


projects/shared/src/lib/components/slider/slider.component.ts
```diff
@@ -126,4 +132,23 @@ export class SliderComponent implements ControlValueAccessor {
     return distPx < avgLabelPx + 16;
   }
 
+  public onInput(event: Event, changeFunction = (v: number) => this.onValueChange(v)): void {
+    if (!this.changeValueWhileSliding) {
+      return;
+    }
+    const input = event.target as HTMLInputElement;
```
steinkobben, line 139: event.target can be null. Can be done without the 'as' operator, maybe something like this:
```suggestion
    if (!(event.target instanceof HTMLInputElement)) {
      return;
    }
    const input: HTMLInputElement = event.target;
```

PR #956, opened 2025-08-07
EK-30: Create info panel

steinkobben:
- [ ] fix menubar panel test


mprins (changes requested):
please use @inject() instead of constructor dependency injection, see https://angular.dev/reference/migrations/inject-function

mprins (approved):

PR #960, opened 2025-08-12
Fix rewrite to new template syntax

matthijsln:
- [x] drop_down icon is suddenly on the left side of the label (when avoiding NG8011 warning???)
- [x] Check other automatic conversion errors which use `else` with `ng-template` which caused original problem


matthijsln:
@geertplaisier heb je enig idee?

geertplaisier (approved):

PR #967, opened 2025-08-20
EO-561: allow saving empty initial values for numeric filters

steinkobben:
[EO-561](https://b3partners.atlassian.net/browse/EO-561) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)

When a numeric filter is saved without initial values, the filter is not applied until a value is entered by the end-user

- [ ] Wait for review from client

[EO-561]: https://b3partners.atlassian.net/browse/EO-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


geertplaisier (changes requested):
When the end-user selects values in the slider (both lower and upper) and then changes the value for lower to 0 - the upper value in the slider is reset to max value

geertplaisier (approved):

PR #1330, opened 2026-05-13
Add DXF export to list of possible exports (not yet supported by API)

geertplaisier:


matthijsln (approved):

mprins:
tailormap-api support for dxf is non-trivial; it would require creating a geotools-dxf-writer module, similar to how we created the excel writing store: https://github.com/Tailormap/gt-excel-writer, but it would much more complex having to write geometric data and loosing attributes makes it non-sensical IMO. 

Note that we cannot use or look at the geoserver code which is **GPL**'ed

PR #1358, opened 2026-06-05
HTM-2040: Feature selection bookmark

steinkobben:
[HTM-2040](https://b3partners.atlassian.net/browse/HTM-2040) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)  

- [x] Test in Iframe
- [x] Remove fragment from bookmark after closing feature info or clicking another feature 

[HTM-2040]: https://b3partners.atlassian.net/browse/HTM-2040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


geertplaisier (changes requested):
In general I think we need to refactor a bit how the services are connected. Right now the MobileLayoutService uses the ApplicationBookmarkService. Since MobileLayoutService is used in various components and ApplicationBookmarkService now needs the API we need to add this to all the tests. 

I think it would be better to nog let the MobileLayoutService depend on ApplicationBookmarkService but let the ApplicationBookmarkService set a value inside the MobileLayoutService (so the other way around). This prevents all other components from requiring API service in tests.

projects/core/src/lib/services/application-bookmark/feature-selection-bookmark.helper.ts
geertplaisier, line 1: Maybe it would be good to write some tests for this file - test the string splitting etc since this could maybe break in certain scenarios

geertplaisier (approved):

PR #1387, opened 2026-06-29
HTM-2080: Follow mode for location

matthijsln:
[HTM-2080](https://b3partners.atlassian.net/browse/HTM-2080) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=Tailormap&utm_campaign=badge_info)  

 

[HTM-2080]: https://b3partners.atlassian.net/browse/HTM-2080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


projects/map/src/lib/openlayers-map/openlayers-map.ts
```diff
@@ -43,10 +43,12 @@ export class OpenLayersMap implements MapViewerModel {
 
   private readonly resizeObserver: ResizeObserver;
   private initialExtent: OpenlayersExtent = [];
-  private initialCenterZoom?: [number[], number] = undefined;
+  private initialCenterZoom?: [ number[], number ] = undefined;
   private mapPadding: number[] | undefined;
 
   private hasUserInteractedSubject = new BehaviorSubject(false);
+  private pointerDragSubject = new Subject<void>();
+  private pointerDragSubscription: Subscription | null = null;
```
geertplaisier: Why add a subscription here? You can use `OpenLayersEventManager.onPointerDrag$()` directly inside the component, I think we use the EventManager in more places directly

geertplaisier (changes requested):

projects/map/src/lib/openlayers-map/openlayers-map.ts
matthijsln: Because if we don't unsubscribe, the second time initMap() is called this would subscribe again. I think that when OpenLayersEventManager.destroy() is called in initMap() the event stream is re-used for the new map so the subscriptions will return events for the new map as well, so lead to a double emit on `pointerDragSubject`? Even if it didn't emit events it would leak a stale subscription.

It's also possible to subscribe once in the constructor instead of initMap instead of unsubscibing.

BTW in OpenLayersEventManager, when initEvents() is called the second time, the skipUntil(OpenLayersEventManager.renderCompleteEvent) for onMapMoveStart$() does not block the events until a render complete event for the new map either.

So maybe a second initMap() call is not really a completely tested/worked out scenario...

geertplaisier (approved):

PR #1467, opened 2026-09-10
Update dependencies to current versions

mprins:


mprins:
checked that printing top PDF with QR code works: [map-10-09-2026__12_18_58.pdf](https://github.com/user-attachments/files/32053941/map-10-09-2026__12_18_58.pdf)

